PR #240, opened 2026-03-28
Update issue templates for new type/label scheme

isaacschepp:
## Summary

- Set native GitHub issue type (Bug/Feature) via `type:` field in templates
- Add required Area dropdown so submitters must pick cli, spec, go-glx, etc.
- Remove references to deleted labels (bug, enhancement, needs-triage)
- Remove `[Bug]`/`[Feature]` title prefixes (types handle this now)
- Disable blank issues via config.yml to ensure templates are used

## Notes

The area dropdown captures the selection in the issue body. Labels still need to be applied manually (or via a future automation workflow). The key improvement is that submitters are forced to think about and declare the area upfront.

### DIFF
--- a/.github/workflows/validate-spec.yml
+++ b/.github/workflows/validate-spec.yml
@@ -13,12 +13,20 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-      
+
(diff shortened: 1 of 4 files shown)


noahbjohnson (approved):
noice

Issue #120, opened 2026-03-10
feat: glx merge — combine two GLX archives


isaacschepp:
## Summary

Add a `glx merge` command that combines two GLX archives into one, with conflict detection and resolution.

## Motivation

Research often happens across multiple archives — one researcher works on the maternal line, another on the paternal line, or you import GEDCOM from different sources. Currently there's no way to combine archives without manual file copying and ID deconfliction.

## Example

```bash
# Merge another archive into the current one
glx merge ./other-archive/ --into ./my-archive/

# Dry run to preview conflicts
glx merge ./other-archive/ --into ./my-archive/ --dry-run

# Auto-resolve by preferring the incoming archive
glx merge ./other-archive/ --into ./my-archive/ --prefer incoming
```

```
Merging ./other-archive/ into ./my-archive/

  New entities:
    12 persons, 8 events, 5 relationships, 3 sources

  Conflicts (same ID, different content):
    person-john-smith — born_on differs: "1850" vs "1852"
    source-census-1860 — title differs

  ID collisions (same ID, unrelated entities):
    event-a3f8d2c1 exists in both archives

  Resolution needed for 3 conflicts. Use --prefer or resolve manually.
```

## Implementation Notes

- Load both archives, detect overlapping entity IDs
- Three-way classification: new (only in source), unchanged (identical), conflicting (same ID, different content)
- ID collision handling: auto-rename with suffix when entities are clearly different
- `--prefer local|incoming` for automatic conflict resolution
- `--dry-run` for preview
- Update all internal references when IDs are remapped
- Related to #94 (multi-researcher workflow) and #98 (glx diff)

PR #264, opened 2026-03-29
feat: Add glx merge command to combine two archives

### LINKED ISSUE #273: Merge produces 197 duplicate vocabulary warnings when both archives have standard vocabs
## Summary

When merging two archives that both embed standard vocabularies, `glx merge` reports ~230 "duplicate" warnings for every standard vocabulary entry. This buries the actually useful information (new entities and real entity ID conflicts).

### Root cause

The `mergeMap` helper in `go-glx/types.go` (line 544) only checks **key existence** — it does NOT compare values:

```go
if _, exists := dest[k]; exists {
    duplicates = append(duplicates, fmt.Sprintf("duplicate %s ID: %s", mapType, k))
}
```

Since both archives load the same standard vocabularies via `LoadStandardVocabulariesIntoGLX()`, every one of the 230 standard vocabulary entries triggers a "duplicate" message even though the values are identical.

### Vocabulary entry counts (230 total)

| Category | Count |
|----------|-------|
| event_types | 39 |
| participant_roles | 30 |
| person_properties | 24 |
| source_types | 19 |
| relationship_types | 17 |
| place_types | 16 |
| media_properties | 16 |
| repository_types | 10 |
| source_properties | 10 |
| repository_properties | 9 |
| citation_properties | 9 |
| media_types | 7 |
| place_properties | 7 |
| event_properties | 5 |
| confidence_levels | 4 |
| gender_types | 4 |
| relationship_properties | 4 |

### CLI UX best practices

Terraform plan is the gold standard: show only what changed, suppress identical items entirely, end with a summary count. The [CLI Guidelines (clig.dev)](https://clig.dev/) say: "The more irrelevant output you produce, the longer it's going to take the user to figure out what they did wrong."

### RDF set semantics as precedent

In RDF, a graph is a **set** of triples — duplicate triples collapse silently by definition. Standard vocabulary entries appearing in both archives are analogous: identical definitions for the same ID should merge silently, only conflicting definitions (same ID, different content) warrant a warning.

### Deep equality in Go

`reflect.DeepEqual` is the pragmatic choice for comparing vocabulary structs. The types are small, comparison happens once per merge, and there are no unexported fields. `google/go-cmp` is for tests only (may panic on unexported fields, performance is not a goal).

---

### Approach: Silent dedup for identical values, report only true conflicts

Modify `mergeMap` to compare values with `reflect.DeepEqual`. Identical entries are silently skipped. Only entries with the **same key but different values** are reported as conflicts. Return the count of identical skips so the CLI can show a summary.

### Changes

**1. `go-glx/types.go` — change `Merge()` return type and `mergeMap` logic**

Change `Merge()` signature from:
```go
func (g *GLXFile) Merge(other *GLXFile) []string
```
to:
```go
func (g *GLXFile) Merge(other *GLXFile) (conflicts []string, identicalSkipped int)
```

Change `mergeMap` to accept a `deepEqual bool` parameter (or create a variant). When `deepEqual` is true and `reflect.DeepEqual(dest[k], src[k])` is true, increment skip counter instead of appending to conflicts:

```go
func mergeMap[V any](mapType string, dest, src map[string]V, deepEqual bool) (conflicts []string, skipped int) {
    for k, v := range src {
        if existing, exists := dest[k]; exists {
            if deepEqual && reflect.DeepEqual(existing, v) {
                skipped++
            } else {
                conflicts = append(conflicts, fmt.Sprintf("duplicate %s ID: %s", mapType, k))
            }
        } else {
            dest[k] = v
        }
    }
    return conflicts, skipped
}
```

Use `deepEqual: false` for entity maps (entity ID collisions are always significant) and `deepEqual: true` for vocabulary/property maps (identical standard vocabs should be silent).

**2. `glx/merge_runner.go` — update output formatting**

Change the duplicates display from listing every entry to a tiered output:

```
  Conflicts (3 — skipped, destination kept):
    duplicate persons ID: person-1
    duplicate events ID: event-1
    duplicate sources ID: source-1

  230 identical vocabulary/property entries skipped
```

When there are no conflicts and only identical skips:
```
  No conflicts. 230 identical vocabulary entries skipped.
```

**3. `go-glx/types_merge_test.go` — update tests**

- Update existing tests to handle the new `(conflicts, skipped)` return values
- Add a test case for identical vocabulary entries (skipped == N, conflicts empty)
- Add a test case for conflicting vocabulary entries (same key, different label → appears in conflicts)

### What NOT to do

- **Don't add `--quiet-vocabs` flag** — silent dedup for identical items is the correct default, not a flag
- **Don't add `--verbose` flag** in this issue — that can be a follow-up for showing full details
- **Don't use `cmp.Equal` from go-cmp** — it's for tests only, may panic, performance is not a goal
- **Don't change entity duplicate behavior** — entity ID collisions are always worth reporting even if values happen to be identical

### Files to modify

1. `go-glx/types.go` — change `Merge()` and `mergeMap()` signatures and logic (~15 lines changed)
2. `go-glx/types_merge_test.go` — update test assertions for new return type (~20 lines changed)
3. `glx/merge_runner.go` — update `mergeArchivesInMemory` and output formatting (~15 lines changed)
4. `glx/merge_runner_test.go` — update test assertions (~5 lines changed)

### LINKED ISSUE #569: glx merge: help text says "entity IDs" but duplicates include vocabularies and metadata
## Summary

The `glx merge` command help text says:

> Duplicate entity IDs are reported and skipped

But the underlying `GLXFile.Merge` also reports/skips duplicates for vocabularies, property definitions, and metadata. When merging two archives with standard vocabularies, this produces ~197 "duplicate" lines that are all vocabulary entries, not entities.

### Fix

Update the help text to:

> Duplicate or conflicting items (entities, vocabularies, property definitions, and metadata) are reported and skipped (destination version kept).

### Related

- #273 (merge produces 197 duplicate vocabulary warnings — the UX side of this issue)

Found during Copilot review of PR #264.

---
Fixed in PR #264.

### LINKED ISSUE #594: merge: handle destPath "." (current directory) for multi-file archives
## Problem

`safeWriteMultiFileArchive` uses `os.Rename(destPath, backupDir)` for its backup-then-swap strategy. When `destPath` is `"."` (the CLI default for `--into`), `os.Rename(".", ...)` fails on all platforms:

- **Linux/macOS**: POSIX [explicitly forbids](https://pubs.opengroup.org/onlinepubs/9799919799/functions/rename.html) renaming "." or ".." — returns EINVAL
- **Windows**: Returns ERROR_SHARING_VIOLATION (errno 32) because the cwd directory has an open handle

### os.Rename and "." behavior

POSIX spec: "If either pathname argument refers to a path whose final component is either dot or dot-dot, rename() shall fail." The rationale is preventing cyclical filesystem paths. This is enforced at the kernel level — Go's `os.Rename` just forwards to the syscall.

**`filepath.Abs(".")` resolves this on POSIX.** Converting "." to its absolute path (e.g., "/home/user/archive") removes the special "." component. The kernel treats it as a normal path and the rename succeeds, even if the process's cwd is that directory (POSIX renames work regardless of open handles/cwds — the inode reference remains valid).

### Windows complication

On Windows, `MoveFileEx` (used by `os.Rename`) **cannot rename a directory that is any process's current working directory**. Even with `filepath.Abs`, if the user runs `cd archive && glx merge ../source/`, both the shell and glx.exe hold the directory as cwd. The rename fails.

Key Windows facts:
- `MoveFileEx` does NOT set `MOVEFILE_COPY_ALLOWED`, so cross-volume moves also fail
- No `renameat2(RENAME_EXCHANGE)` equivalent exists on Windows
- Windows Transactional NTFS (TxF) is deprecated by Microsoft

### Go 1.24-1.26 filesystem APIs

- **Go 1.24**: Added `os.Root` + `os.OpenRoot()` for sandboxed filesystem access
- **Go 1.25**: Added `os.Root.Rename()` for sandboxed renames
- **Go 1.26**: No `os.Rename` changes, no `renameat2` wrapper, no atomic directory swap

No standard library solution for atomic directory replacement exists.

### Production patterns (from GitHub research)

| Project | Pattern | Source |
|---------|---------|--------|
| [OpenShift atomicdir](https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/internal/atomicdir/sync.go) | Write to staging → `renameat2(RENAME_EXCHANGE)` → cleanup old | Linux only |
| [Canonical snapd](https://github.com/canonical/snapd/blob/master/osutil/rename_linux.go) | `renameat2(RENAME_EXCHANGE)` with absolute paths | Linux only |
| [Dolt database](https://github.com/dolthub/dolt/blob/main/go/libraries/utils/file/sync.go) | Write to temp → fsync file → rename → fsync parent dir | Cross-platform |
| [gowarc](https://github.com/nlnwa/gowarc/blob/master/warcfile.go) | `os.Rename` + `fsync` parent directory for crash safety | Cross-platform |
| [cattlecloud/atomicfs](https://github.com/cattlecloud/atomicfs/blob/main/atomicfs.go) | Verify same-device → write to temp → fsync → rename → fsync parent | Cross-platform |

**Crash-safety best practice**: After `os.Rename`, `fsync` the parent directory. Without it, rename metadata may not be persisted (the Dolt and gowarc patterns both do this).

### filepath.Abs reliability

- Reliable on all platforms for normal use
- Does NOT resolve symlinks (use `filepath.EvalSymlinks` if needed)
- Go caches `os.Getwd()` — invalidated by `os.Chdir()` but NOT by external renames of parent dirs

---

### Approach: filepath.Abs + Windows error handling

A one-line fix (`filepath.Abs`) resolves the issue on Linux/macOS. On Windows, if the rename fails due to the cwd lock, return a clear error message.

### Changes

**1. `mergeArchives()` in `merge_runner.go` — resolve paths to absolute early**

```go
func mergeArchives(srcPath, destPath string, dryRun bool) error {
    // Resolve to absolute paths so "." becomes a real path that can be renamed.
    var err error
    destPath, err = filepath.Abs(destPath)
    if err != nil {
        return fmt.Errorf("resolving destination path: %w", err)
    }
    srcPath, err = filepath.Abs(srcPath)
    if err != nil {
        return fmt.Errorf("resolving source path: %w", err)
    }

    // ... rest of function unchanged ...
}
```

This is the primary fix. By converting "." to "/absolute/path/to/archive" before it reaches `safeWriteMultiFileArchive`, the `os.Rename` calls operate on normal paths.

**2. Fix the misleading comment in `mergeArchives()`**

The comment on line 192-194 currently says:
```go
// Save — multi-file archives use crash-safe temp+swap to prevent destruction
// if the write fails. Single-file archives are written directly (atomic rename
// of a single file is handled by the OS).
```

Replace with:
```go
// Save — multi-file archives use crash-safe temp+swap to prevent partial writes.
// Single-file archives use os.WriteFile (not atomic; see #595 for improvement).
```

**3. Add a test for "." as destination**

```go
func TestMergeArchives_DotDestination(t *testing.T) {
    // Create dest archive in temp dir
    destDir := t.TempDir()
    // ... write a person to destDir ...

    // Create source archive in separate temp dir
    srcDir := t.TempDir()
    // ... write a different person to srcDir ...

    // Save original cwd, chdir into dest, merge with "."
    origDir, err := os.Getwd()
    require.NoError(t, err)
    defer os.Chdir(origDir)

    require.NoError(t, os.Chdir(destDir))
    err = mergeArchives(srcDir, ".", false)
    require.NoError(t, err)

    // Verify merge result
    reloaded, dupes, err := LoadArchiveWithOptions(destDir, false)
    require.NoError(t, err)
    assert.Empty(t, dupes)
    assert.Len(t, reloaded.Persons, 2)
}
```

**4. (Optional) Add fsync of parent directory after rename for crash safety**

Per the Dolt/gowarc patterns, add fsync after the rename in `safeWriteMultiFileArchive`:

```go
// Move temp into place
if err := os.Rename(tmpDir, destPath); err != nil {
    _ = os.Rename(backupDir, destPath)
    return fmt.Errorf("moving merged archive into place: %w", err)
}

// Sync parent directory to ensure rename is persisted to disk
if pdir, err := os.Open(filepath.Dir(destPath)); err == nil {
    _ = pdir.Sync()
    _ = pdir.Close()
}
```

### What NOT to do

- **Don't add `renameat2`/`RENAME_EXCHANGE`**: Linux-only, requires `golang.org/x/sys/unix` dependency, and the current backup+swap pattern is sufficient for a CLI tool
- **Don't add `os.Chdir()` to work around Windows**: Changing the process cwd has side effects and the user's shell still holds the directory. Better to fail with a clear message.
- **Don't switch to in-place update**: Loses crash safety that `safeWriteMultiFileArchive` was specifically designed for

### Windows behavior after fix

With `filepath.Abs`, the rename uses an absolute path. If the user is cd'd into the destination:
- The Go process has inherited that cwd
- `os.Rename` on the absolute path will fail with ERROR_SHARING_VIOLATION
- The existing error handling catches this: `return fmt.Errorf("backing up original: %w", err)`
- The error message is already clear enough: "backing up original: rename C:\Users\...\archive C:\Users\...\archive.bak: The process cannot access the file because it is being used by another process."

This is acceptable — the user just needs to run from a parent directory or use `--into path/to/archive` instead of relying on the "." default.

### Relationship to other issues

- **#595** (single-file not atomic): The comment fix in step 2 addresses the misleading claim. The actual atomic write fix is separate scope.
- **#593** (media binary copying): Unrelated to path resolution but affects the same `safeWriteMultiFileArchive` function.
- **#570** (dry-run test): Orthogonal — different test case.

### Files to modify

1. `glx/merge_runner.go` — add `filepath.Abs` calls + fix comment
2. `glx/merge_runner_test.go` — add "." destination test

### LINKED ISSUE #595: merge: writeSingleFileArchive uses os.WriteFile, not atomic temp+rename
## Problem

`writeSingleFileArchive` uses `os.WriteFile` which opens the file with `O_TRUNC`, immediately destroying the old content before writing new content. A failed write (disk full, power loss, signal) leaves a corrupt or zero-length file. The merge command's comment previously claimed this was "atomic" but it is not.

The misleading comment was already fixed in commit 9ab453a (#594).

### os.WriteFile internals

`os.WriteFile` opens with `O_WRONLY|O_CREATE|O_TRUNC`, which **truncates the file to zero bytes before any data is written**. On failure at any point after open, the old content is permanently lost.

### Go standard library status

- No atomic file write utility exists in Go 1.24-1.26
- [Proposal #22397](https://github.com/golang/go/issues/22397) to add one has been open since 2017 with no progress
- [Issue #56173](https://github.com/golang/go/issues/56173) was merged to document that `os.WriteFile` is not atomic

### Available libraries

| Library | Stars | Windows? | fsync? | API | Notes |
|---------|-------|----------|--------|-----|-------|
| [google/renameio/v2](https://github.com/google/renameio) | 671 | No | File only | `WriteFile()` + `NewPendingFile()` | Most mature; Linux/macOS only |
| [natefinch/atomic](https://github.com/natefinch/atomic) | 208 | Yes | File + WRITE_THROUGH | `WriteFile(filename, io.Reader)` | Cross-platform; uses deprecated ioutil |
| [moby/sys/atomicwriter](https://pkg.go.dev/github.com/moby/sys/atomicwriter) | New | Yes | Yes | `WriteFile()` drop-in for os.WriteFile; also `WriteSet` for multi-file | Newest (April 2025), from Docker team |

### The correct atomic write pattern

```
1. os.CreateTemp(filepath.Dir(target), ".tmp-*")   — same directory = same filesystem
2. Write content to temp file
3. f.Sync()                                         — ensures data on stable storage
4. f.Close()
5. os.Chmod(temp, filePermissions)                  — set correct permissions before rename
6. os.Rename(temp, target)                          — atomic on POSIX (same filesystem)
7. (optional) fsync parent directory                — for crash durability beyond atomicity
```

On failure at any step, remove the temp file. The original file remains untouched.

### fsync considerations

- **File fsync** (step 3): Prevents a crash from leaving a 0-length file at the target after rename. Required for atomicity.
- **Directory fsync** (step 7): Prevents the rename itself from being rolled back after a power loss. Required for full durability, but [google/renameio explicitly skips it](https://github.com/google/renameio/issues/11) for performance, arguing that atomicity (no corrupt files) is sufficient for most applications.
- For a genealogy CLI tool, file fsync is sufficient. Directory fsync is overkill.

### Windows behavior

- `os.Rename` uses `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING` — works for files (unlike directories)
- Not guaranteed atomic by MSDN, but atomic in practice on NTFS for same-volume files
- `natefinch/atomic` adds `MOVEFILE_WRITE_THROUGH` for extra safety
- Files opened by Go can be renamed over (Go sets `FILE_SHARE_DELETE`), but files held by other programs (editors, antivirus) may block the rename

### Which callers are affected

| Caller | Context | Risk |
|--------|---------|------|
| `mergeArchives()` single-file path | **Overwrites existing file** | HIGH — this is the primary target of this issue |
| `runMigrate()` single-file path | Overwrites existing file | HIGH — tracked in #597 |
| `renameEntities()` single-file path | Overwrites existing file | HIGH — tracked in #597 |
| `importToSingleFile()` | Creates new file | LOW — no existing file to corrupt |
| `joinArchive()` | Creates new file (checks existence first) | LOW |

---

### Approach: Add `atomicWriteFile` helper, no new dependencies

Roll a minimal atomic write function (~25 lines) rather than adding a library dependency. The pattern is simple, well-understood, and avoids pulling in a dependency for one function.

### Changes

**1. Add `atomicWriteFile()` to `fileops.go`**

```go
// atomicWriteFile writes data to a file atomically using a temp file + rename.
// The target file either has the old content or the new content, never a partial
// write. The temp file is created in the same directory to ensure same-filesystem
// rename.
func atomicWriteFile(path string, data []byte, perm os.FileMode) error {
    dir := filepath.Dir(path)
    tmp, err := os.CreateTemp(dir, ".glx-tmp-*")
    if err != nil {
        return fmt.Errorf("creating temp file: %w", err)
    }
    tmpPath := tmp.Name()

    // Clean up temp file on any failure
    success := false
    defer func() {
        if !success {
            tmp.Close()
            os.Remove(tmpPath)
        }
    }()

    if _, err := tmp.Write(data); err != nil {
        return fmt.Errorf("writing temp file: %w", err)
    }
    if err := tmp.Sync(); err != nil {
        return fmt.Errorf("syncing temp file: %w", err)
    }
    if err := tmp.Close(); err != nil {
        return fmt.Errorf("closing temp file: %w", err)
    }
    if err := os.Chmod(tmpPath, perm); err != nil {
        return fmt.Errorf("setting file permissions: %w", err)
    }
    if err := os.Rename(tmpPath, path); err != nil {
        return fmt.Errorf("replacing target file: %w", err)
    }

    success = true
    return nil
}
```

**2. Update `writeSingleFileArchive()` in `archive_io.go`**

Replace `os.WriteFile` with `atomicWriteFile` only when overwriting an existing file. For new files, `os.WriteFile` is fine (no old content to corrupt).

The simplest approach: always use `atomicWriteFile` since it works for both new and existing files (rename creates the file if it doesn't exist):

```go
func writeSingleFileArchive(path string, glx *glxlib.GLXFile, validate bool) error {
    serializer := createSerializer(validate, true, "  ")
    yamlBytes, err := serializer.SerializeSingleFileBytes(glx)
    if err != nil {
        return fmt.Errorf("failed to serialize GLX file: %w", err)
    }
    if err := atomicWriteFile(path, yamlBytes, filePermissions); err != nil {
        return fmt.Errorf("failed to write GLX file: %w", err)
    }
    return nil
}
```

**3. Add test for atomic write**

```go
func TestAtomicWriteFile(t *testing.T) {
    dir := t.TempDir()
    path := filepath.Join(dir, "test.glx")

    // Write initial content
    require.NoError(t, atomicWriteFile(path, []byte("original"), 0o644))
    data, err := os.ReadFile(path)
    require.NoError(t, err)
    assert.Equal(t, "original", string(data))

    // Overwrite atomically
    require.NoError(t, atomicWriteFile(path, []byte("updated"), 0o644))
    data, err = os.ReadFile(path)
    require.NoError(t, err)
    assert.Equal(t, "updated", string(data))

    // Verify permissions
    info, err := os.Stat(path)
    require.NoError(t, err)
    assert.Equal(t, os.FileMode(0o644), info.Mode().Perm())
}

func TestAtomicWriteFile_NewFile(t *testing.T) {
    dir := t.TempDir()
    path := filepath.Join(dir, "new.glx")

    require.NoError(t, atomicWriteFile(path, []byte("content"), 0o644))
    data, err := os.ReadFile(path)
    require.NoError(t, err)
    assert.Equal(t, "content", string(data))
}
```

### What NOT to do

- **Don't add a library dependency** (renameio, natefinch/atomic, moby/sys) — the pattern is 25 lines, well-understood, and adding a dependency for one function is overkill for a CLI tool
- **Don't fsync the parent directory** — atomicity (no corrupt files) is sufficient; full durability (survive power-loss rename rollback) is overkill for a CLI tool whose archives are in Git
- **Don't change callers that write to NEW files** (import to new path, join to new path, init) — no corruption risk when there's no existing file to lose
- **Don't change the multi-file write path** — `safeWriteMultiFileArchive` already handles crash safety for multi-file merges via temp+swap

### Files to modify

1. `glx/fileops.go` — add `atomicWriteFile()`
2. `glx/archive_io.go` — change `writeSingleFileArchive()` to use `atomicWriteFile()`
3. `glx/fileops_test.go` (or new test file) — add tests for `atomicWriteFile()`

### Relationship to other issues

- **#594** (dot path): Already fixed the misleading comment referencing this issue
- **#597** (migrate/rename crash safety): Filed during this research — those commands should also adopt `atomicWriteFile` once it exists

isaacschepp:
## What and why

New `glx merge` command that combines two GLX archives, with duplicate detection and dry-run support. This implements the basic merge flow from #120; conflict resolution (`--prefer`) is deferred to a follow-up.

```bash
glx merge ./other-archive/ --into ./my-archive/
glx merge ./other-archive/ --into ./my-archive/ --dry-run
```

## Related issues

Partial implementation of #120 (basic merge with duplicate skip; `--prefer` conflict resolution deferred)
Closes #273
Closes #569
Closes #594
Closes #595

## Testing

- [x] 3 unit tests: new entities, duplicates, empty source
- [x] 1 integration test: disk round-trip (write, merge, reload, verify no duplicates)
- [x] 1 test for "." as destination (chdir into dest, merge with ".", verify result)
- [x] 5 tests for atomicWriteFile (new file, overwrite, permissions, no temp leak, invalid dir)
- [x] 2 tests for identical vocab dedup (silently skipped) and conflicting vocab (reported)
- [x] 24 library-level merge tests pass
- [x] Full test suite passes, `go vet` clean

## Breaking changes

None


noahbjohnson (changes requested):
A quick docs pass in these places is the only blocker:
  - CHANGELOG.md — no entry for glx merge
  - glx/README.md — not updated
  - website/.vitepress/config.js — not updated
  - docs/guides/hands-on-cli-guide.md — not updated

noahbjohnson (approved):

Issue #388, opened 2026-03-30
CHANGELOG: use [Unreleased] header instead of pre-committed version number


isaacschepp:
## Problem

The unreleased section uses the next version number:

```markdown
## [0.0.0-beta.10]
```

Keep a Changelog specifies `[Unreleased]` for changes not yet released:

```markdown
## [Unreleased]
```

Pre-committing the version number is problematic: it may change before release, and implies the version exists.

## Proposed Change

Rename `## [0.0.0-beta.10]` to `## [Unreleased]` and update the comparison link.

At release time, rename to the actual version with date, then add a new empty `[Unreleased]` section above.

## Why

- Matches Keep a Changelog specification exactly
- Avoids pre-committing version numbers that may change
- Makes it clear these changes are not yet released

PR #585, opened 2026-03-31
fix: Preserve place-less GEDCOM RESI records as residence events

### LINKED ISSUE #488: GEDCOM import: RESI records with no PLAC lost (place-reference model limitation)
## Summary

RESI records with only DATE/TYPE/value but no PLAC sub-record cannot be stored in the place-reference residence model. 7 such records in the bullinger test file are bare markers like `RESI Y` or `RESI\n2 TYPE married`.

### Evidence
- bullinger: RESI 26→19 (7 lost)

_Migrated from todo.md — Model Limitations_

isaacschepp:
## What and why

GEDCOM RESI records without a PLAC sub-record were silently dropped during import. The bullinger test file lost 7 RESI records (26 imported → 19 survived). These include bare `RESI Y` markers and RESI with only DATE or TYPE — valid GEDCOM data that was being discarded.

### Root cause

The `residence` person property has `reference_type: places`, so its value must be a Place entity ID. When no PLAC exists, there's no Place ID to store, and `convertResidence` exited early.

### The fix

When a RESI has no PLAC, create a `residence` Event entity instead of a person property. This preserves date, TYPE, and notes that would otherwise be lost.

- RESI with PLAC → person property (unchanged)
- RESI without PLAC → residence Event entity (new)

Added `residence` to the standard event types vocabulary with `gedcom: "RESI"`.

## Related issues

Fixes #488

## Testing

- `TestConvertResidence_NoPlacCreatesEvent` — RESI with DATE+TYPE, no PLAC → creates event with date
- `TestConvertResidence_BareRESIYCreatesEvent` — bare `RESI Y` → creates event (not dropped)
- All existing tests pass (E2E bullinger, maximal, roundtrip)

## Breaking changes

None. Existing archives are unaffected. Previously-dropped RESI records will now appear as events on re-import.

### DIFF
--- a/.github/workflows/auto-resolve-conflicts.yml
+++ b/.github/workflows/auto-resolve-conflicts.yml
@@ -96,6 +96,27 @@ jobs:
             # Resolve by keeping both sides (remove conflict markers)
             sed -i '/^<<<<<<< /d; /^=======$/d; /^>>>>>>> /d' CHANGELOG.md
 
+            # Deduplicate section headers produced by keep-both-sides merge.
+            # When both sides add entries under the same ### or #### heading,
+            # the resolved file contains duplicate headers. This awk pass
+            # removes the second occurrence within each release section,
+            # resetting at each ## [...] boundary to avoid cross-release
+            # collapsing.
+            awk '
+              /^## \[/ { delete seen }
+              /^#{3,4} / {
+                if (seen[$0]) { blank_pending = 0; next }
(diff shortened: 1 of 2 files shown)


noahbjohnson (changes requested):
couple nits and needs a changelog entry (use unreleased and we can close #388 as well)

would also like to know current actual roundtrip behavior here to know if we need a follow-up there

PR #691, opened 2026-04-13
fix: Mark line-ending-sensitive GEDCOM fixtures -text

noahbjohnson:
## What and why

The `v0.0.0-beta.10` release workflow [failed](https://github.com/genealogix/glx/actions/runs/24360638945) with:

```
⨯ release failed after 0s
  │ git is in a dirty state
  │ Please check in your pipeline what can be changing the following files:
  │  M glx/testdata/gedcom/5.5.1/...
  │  M glx/testdata/gedcom/5.5.5/...
  │  (20 files)
```

**Root cause**: `.gitattributes` has `* text=auto` which tells git to normalize line endings to LF. PR #656 renormalized most of the repo but [deliberately reverted](https://github.com/genealogix/glx/commit/220105c) the `glx/testdata/gedcom/` tree "to avoid changing test fixtures that may be sensitive to line endings" — without also adding a `-text` attribute to exclude them from `text=auto`. Git therefore kept reporting those files as dirty (working tree has CRLF, `text=auto` wants LF), which GoReleaser treats as a fatal error.

This went unnoticed on v0.0.0-beta.9 because that release was cut before #656 landed.

**Fix**:
- Add `-text` rules in `.gitattributes` for fixtures whose line endings genuinely matter:
  - `glx/testdata/gedcom/5.5.5/**/*.ged` — GEDCOM 5.5.5 spec mandates CRLF + UTF-8 BOM
  - `glx/testdata/gedcom/5.5.1/torture-test-551/*.ged` — files named `*LF.ged` / `*CLF.ged` are explicit line-ending variants for torture testing
  - Existing `cp1252-crlf.ged` rule kept
- Renormalize the rest to LF: 13 non-line-ending-sensitive `.ged` fixtures in 5.5.1 and 18 `testdata/gedcom` README.md files

## Related issues

Fixes the failed release run on tag `v0.0.0-beta.10`. No open issue — failure surfaced post-tag.

## Testing

- `git status` now clean on the tip of this branch (both locally and verified that the renormalized blobs are what fresh checkout produces)
- `make test` — all packages pass (`glx`, `go-glx`)
- `git check-attr text -- <fixture>` confirms the protected fixtures report `text: unset`

## Breaking changes

None. Line-ending-sensitive fixtures still have their original bytes (now explicitly preserved). Non-sensitive fixtures now have LF, which matches repo convention.

## Follow-up

After merge, the `v0.0.0-beta.10` tag needs to be deleted and re-created from the merge commit so the release workflow re-runs with a clean tree.


isaacschepp (approved):
Clean fix. Root cause is well identified and the broad `glx/testdata/gedcom/**/*.ged -text` glob is the right call — simpler and more defensive than per-directory rules, and prevents this class of issue from recurring when new fixtures are added.

Minor note: the PR body describes fine-grained per-directory `-text` rules (5.5.5, torture-test-551), but the actual implementation uses the single broad glob — which is better. Just a description/implementation mismatch, no action needed.

README renormalizations look correct (additions == deletions across all of them), CHANGELOG entry is there, all CI green. LGTM.

Re: follow-up — once this merges, the `v0.0.0-beta.10` tag needs to be deleted and re-created from the merge commit. Happy to help with that or leave it to you.

Issue #702, opened 2026-04-14
Add --preview flag to glx merge for dry-run analysis


isaacschepp:
## Area

cli

## Summary

Add a `--preview` (or `--dry-run`) flag to `glx merge` that shows what would happen without modifying any files. This is the prerequisite for safe multi-researcher collaboration — researchers need to see the impact of a merge before committing to it.

## Proposed output

```
$ glx merge --preview research/brown-family/ main-archive/

Merge preview: research/brown-family/ → main-archive/

  New entities (3):
    person-jane-brown        (Jane Brown, b. 1832)
    event-birth-jane-brown   (birth)
    assertion-jane-birth     (birth date assertion)

  Conflicts (1):
    person-robert-webb
    → properties.birth_date: "1815" (dest) vs "ABT 1815" (src)

  Potential duplicates (1):
    person-jane-miller (src) ↔ person-jane-webb (dest)
    Names: "Jane Miller" / "Jane Webb"  Score: 0.85

  Vocabulary: 0 conflicts, 12 identical skipped
```

## Implementation

The existing `mergeArchives()` in `merge_runner.go` already computes conflicts and new entity counts. Add:
1. A `--preview` flag to the merge cobra command
2. Before calling `dest.Merge(src)`, run `glx.DuplicateDetection()` across both archives to find potential duplicates
3. Report the merge result without writing to disk
4. Include potential duplicates in the report (leveraging existing `go-glx/duplicates.go` scoring)

## Parent issue

- #94 (Multi-researcher workflow)

Issue #94, opened 2026-03-09
feat: multi-researcher workflow — collaboration docs and merge tooling


isaacschepp:
## Summary

Documentation and tooling for multiple researchers collaborating on a shared GLX archive via Git. Covers branch strategies, genealogy-specific merge conflict resolution, and workflows for the common case where two researchers independently add evidence for the same person or family.

## Motivation

Genealogy is inherently collaborative. Researchers share findings, divide work by family line or record type, and reconcile conflicting evidence. GLX's plain-text format and Git backend make collaboration technically possible today, but there's no guidance or tooling for:

- How to structure branches for parallel research
- What happens when two researchers add the same person independently
- How to resolve conflicts when two assertions disagree on a fact
- How to merge evidence from two separate archives into one

Without this, multi-researcher workflows will produce messy merge conflicts in YAML files that Git's line-level merge can't handle intelligently.

#### By family line
```
main
├── research/webb-family        (Researcher A)
├── research/brown-family       (Researcher B)
└── research/hartford-county-census (Researcher C)
```
Low conflict risk — different persons and files. Merge to main when research is solid.

#### By record type
```
main
├── research/census-1860
├── research/vital-records
└── research/military-records
```
Higher conflict risk — multiple branches may touch the same person. Requires merge tooling.

#### By confidence level
```
main                            (reviewed, high-confidence)
├── draft/jane-webb-parents     (speculative research)
└── draft/robert-webb-military       (in-progress sourcing)
```
Draft branches hold work-in-progress. Promote to main via PR when evidence meets the standard.

### Collaboration patterns

Document common scenarios:
1. **Divide and conquer** — each researcher owns a family line, periodic merges
2. **Source specialist** — one researcher transcribes records, another analyzes and creates assertions
3. **Review workflow** — one researcher drafts, another reviews evidence quality before merge
4. **Archive joining** — two independent archives for the same family need to be combined

### Genealogy-aware conflict resolution

Git's default merge operates at the line level. GLX needs entity-level awareness:

```
glx merge <branch>              # Merge branch into current, with GLX-aware resolution
glx merge --preview <branch>    # Show what would conflict without merging
glx merge --archive other/      # Merge a separate archive directory (not Git)
```

#### 1. Duplicate persons (most common)
Two researchers independently create person records for the same individual.

```
# Branch A created:  persons/person-a1b2c3d4.glx  (Robert Webb)
# Branch B created:  persons/person-e5f6g7h8.glx  (R Webb)
```

[...]

PR #761, opened 2026-05-01
feat: Add possibly_same_person relationship type

### LINKED ISSUE #227: feat: Uncertain person identification linking
## Problem
When working through records from a single place, researchers frequently encounter people who MAY be the same individual across different records but cannot be confirmed. GLX has no formal way to express "person A in record X might be the same as person B in record Y."

## Current workaround
Using `notes` fields to document suspicions, but this is not structured or queryable.

## Proposed solution
Either:
1. A new relationship type `possibly_same_person` with confidence level, or
2. A new assertion type that links two person entities as potential matches

This would allow researchers to track uncertain identifications and revisit them when new evidence emerges.

isaacschepp:
## What and why

Adds a new standard relationship type `possibly_same_person` so researchers
can structurally express "person A in record X **might** be the same as
person B in record Y" — the case [issue #227](https://github.com/genealogix/glx/issues/227)
describes, where the current workaround is unstructured `notes` fields.

The issue offered two options: (1) a new relationship type with confidence
or (2) a new assertion type. This PR takes option 1 because confidence is
already a first-class field on `Assertion`, and assertions can already take
a relationship as their `subject`. So adding the vocabulary entry plus a
documented usage pattern (relationship + companion assertion-with-confidence
+ citations) is the minimum-viable structured representation. No JSON
schema changes, no Go type changes.

## Changes

- `specification/5-standard-vocabularies/relationship-types.glx` — new vocabulary entry (single-line `description`, peer style — no `gedcom:` field, matching `step_parent` / `godparent` / `guardian` / `partner` / `relative`)
- `specification/4-entity-types/relationship.md` — new "Possibly Same Person" usage section (worked example pairs the relationship with an Assertion carrying `confidence`/`status`/citations) and a row in the GEDCOM Mapping table
- `specification/4-entity-types/vocabularies.md` — prose summary updated for discoverability
- `CHANGELOG.md` — `### Added → #### Specification` entry

## Related issues

Closes #227

## Testing

- `node specification/validate-schemas.mjs` — all schemas pass
- `bash scripts/check-links.sh` — 337 internal links pass (including the two new ones to `assertion#confidence` and `vocabularies#confidence-levels-vocabulary`)
- `go test ./...` — no new failures (one pre-existing failure on `origin/main` in `TestSerializeMultiFileCrossTypeNoCollision` is unrelated to this change; verified by re-running with my changes stashed)
- Manual end-to-end smoke test: built the CLI, ran `./bin/glx init` against a scratch dir, authored a `possibly_same_person` relationship plus a companion assertion with `confidence: medium`, citations, and `status: speculative`, then ran `./bin/glx validate` — passes cleanly. A typo (`possibly_same_persn`) is correctly rejected with `references non-existent relationship_types`.

## Breaking changes

None. Pure addition: new vocabulary key + new docs section. Existing archives are unaffected; existing relationship types are untouched. No GEDCOM importer/exporter changes (intentional — see GEDCOM Mapping note in `relationship.md`).

## Out of scope (explicit)

- GEDCOM `ALIA` import/export mapping — `ALIA` asserts identity (stronger than "possibly"), so the semantics don't line up. A future PR could export a `possibly_same_person` relationship as `ALIA` only when its companion assertion has `confidence: high` and `status: proven`, but that requires deciding on cross-entity export logic and is its own design discussion.
- A `glx detect-duplicates`-style CLI — already covered by `glx merge --preview` (#94, #702).
- A new participant-role pair like `same_person_candidate` — the relationship is symmetric (both participants are the same kind of "candidate"), so roles are noise and are deliberately omitted, matching `housemate` / `neighbor` / `coworker` / `relative`.


noahbjohnson (approved):
damn, this is a pretty elegant solution. ⭐ 

will worry about more docs in the release prep step

PR #938, opened 2026-05-28
ci(release): pin cosign-installer to v4.1.2 (no floating v4 tag)

noahbjohnson:
## Summary

- Previous bump (#937) used `sigstore/cosign-installer@v4`, but the action only publishes explicit `vX.Y.Z` tags — there's no floating `v4` major-version tag.
- The release workflow on the moved `v0.0.0-beta.11` tag failed at 'Set up job' with `Unable to resolve action sigstore/cosign-installer@v4, unable to find version v4`.
- Same lesson as #779 (`ossf/scorecard-action` for the identical reason).
- Pinned to `v4.1.2` (2026-05-07) — the v4-line release that added cosign 3.0.6 bundle-format support.

## Test plan

- [ ] After merge, re-tag `v0.0.0-beta.11` at the new main HEAD and confirm goreleaser uploads binaries to the existing release.

### DIFF
--- a/.github/CLAUDE.md
+++ b/.github/CLAUDE.md
@@ -1,0 +1,82 @@
+# .github/ — Claude Guide
+
+## Third-party GitHub Actions: always pin to `vX.Y.Z`, never `@vN`
+
+**Verify before you push.** Most third-party actions in this repo bumped via
+`@vN` floating tags have broken at some point because the upstream maintainer
+doesn't publish a floating major-version tag. Symptoms:
+
+- `Unable to resolve action <owner>/<repo>@vN, unable to find version vN`
+  (action only ships `vX.Y.Z`, no floating major)
+- A release run that pulls a stale binary path because the floating tag never
+  moved when a new minor shipped
+
+Before editing any `uses:` line in `.github/workflows/*.yml`, **check what tags
+the action actually publishes**:
+
+```bash
+gh api repos/<owner>/<repo>/tags --jq '.[].name' | head
+```
+
+If you only see `v4.1.2`, `v4.1.1`, etc. and no bare `v4`, pin to the latest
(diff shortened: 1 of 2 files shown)


isaacschepp:
Minor cross-ref cleanup spotted during a `.github/CLAUDE.md` review: the inline comment on the cosign-installer pin at `release.yml:42` cites **#779** (`see #779 for the same lesson on scorecard-action`). #779 is the *scorecard-action* precedent; the cosign-installer pin is **this** issue (#938).

The `.github/CLAUDE.md` "Known instances" table correctly attributes cosign-installer to #938, so the workflow comment is the one to reconcile — point it at #938 for cosign and keep #779 as the scorecard precedent. Folded into #1046's doc rewrite so both land together.

isaacschepp:
**Superseded by #1263 / #1264.** `sigstore/cosign-installer` is no longer on the exact patch tag `@v4.1.2`; it is now SHA-pinned like every other action:

```yaml
uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2 -- only vX.Y.Z tags published, no floating v4; do not "tidy" to @v4 (#938)
```

The finding this issue recorded — that the action publishes no floating `v4`, so `@v4` fails with `Unable to resolve action` — is still true and is preserved as the trailing note on the `uses:` line. What changed is that it no longer needs a pin-form carve-out: a SHA is reachable whether or not a major tag exists, so the "no floating major" case and the ordinary case now take the same form. The patch tag was itself a mutable ref, which was the weakest link in the old three-tier convention.

This also cleared Code Scanning `Pinned-Dependencies` alert 208 (`release.yml:71`), the one alert in that set that wasn't a first-party float.

PR #977, opened 2026-06-03
ci: Validate examples in a parallel matrix instead of a shell loop

### LINKED ISSUE #352: validate-examples: replace shell loop with matrix strategy
## Problem

The `validate-examples` job uses a shell for-loop to validate all example archives:

```yaml
- name: Validate All Examples
  run: |
    for dir in docs/examples/*/; do
      echo "Validating $dir"
      ./bin/glx validate "$dir"
    done
```

Issues with this approach:

1. **Sequential execution**: Each example is validated one after another. As the example count grows, this becomes slower.
2. **Poor error attribution**: If validation fails, the GitHub Actions UI shows the entire step as failed. You must scroll through the log to find which specific example failed  especially painful when multiple examples fail.
3. **First-failure stops**: Due to `set -e` (default bash behavior in Actions), the loop exits on the first failure and skips remaining examples. You don't see the full picture of what's broken.
4. **No parallelism**: Examples are independent and could be validated concurrently.

### Option A: Matrix strategy (preferred)

Use a matrix to validate each example as a separate job step:

```yaml
validate-examples:
  runs-on: ubuntu-latest
  timeout-minutes: 10
  strategy:
    fail-fast: false  # Run all examples even if one fails
    matrix:
      example:
        - minimal
        - basic-family
        - complete-family
        - single-file
        - assertion-workflow
        - participant-assertions
        - temporal-properties
        - westeros
  steps:
    - uses: actions/checkout@v6

    - name: Set up Go
      uses: actions/setup-go@v6
      with:
        go-version: '1.26'

    - name: Build glx CLI
      run: go build -o bin/glx ./glx

    - name: Validate ${{ matrix.example }}
      run: ./bin/glx validate docs/examples/${{ matrix.example }}/
```

**Pros**: Each example gets its own job with clear pass/fail status in the GitHub UI. All examples run even if one fails. Runs in parallel.

**Cons**: More runner time (each matrix entry checks out + builds separately). Could be mitigated by uploading the binary as an artifact from a build job and downloading in matrix jobs.

### Option B: Improved shell script with summary

If matrix overhead is undesirable:

```yaml
- name: Validate All Examples
  run: |
    FAILED=0
    for dir in docs/examples/*/; do
      name=$(basename "$dir")
      echo "::group::Validating $name"
      if ./bin/glx validate "$dir"; then
        echo "::endgroup::"
      else
        echo "::endgroup::"
        echo "::error::Validation failed for $name"
        FAILED=1
      fi
    done
    exit $FAILED
```

This uses `::group::`/`::endgroup::` for collapsible output and `::error::` annotations, runs all examples even on failure, and reports a summary.

## Why

- As the project grows (currently 8 examples), validation time and diagnostic difficulty increase linearly
- Matrix strategy is the idiomatic GitHub Actions pattern for this use case
- Clear per-example pass/fail in the UI reduces debugging time
- `fail-fast: false` gives a complete picture of what is broken

isaacschepp:
## What and why

The `validate-examples` job in `validate-spec.yml` validated every archive under
`docs/examples/*/` with a sequential shell `for` loop:

```yaml
for dir in docs/examples/*/; do
  echo "Validating $dir"
  ./bin/glx validate "$dir"
done
```

As #352 notes, this runs serially, stops at the first failure (default `set -e`),
and reports every example under one opaque step — so a failure means scrolling the
log to find which example broke, and you never see the rest.

This PR implements the issue's preferred **Option A (matrix strategy)** with the
artifact mitigation the issue calls out, plus two refinements:

- **`build-cli`** — builds the CLI **once**, uploads `bin/glx` as an artifact
  (so matrix legs don't each rebuild — the issue's stated mitigation for matrix
  runner-time cost), and emits the example list as a **dynamic matrix** discovered
  with `find docs/examples … | jq`. This preserves the old loop's auto-discovery:
  a newly added example directory is validated with **no workflow edit** (a
  hardcoded matrix list would silently skip new examples — a regression the
  current loop doesn't have).
- **`validate-example`** — matrix job, one example per leg, `fail-fast: false`,
  downloads the prebuilt binary. Gives **per-example pass/fail** in the Actions
  UI and runs **in parallel**.
- **`validate-examples`** — a tiny aggregator job that reports a single pass/fail.

### Why the aggregator job

The repo ruleset (`Main Protection`) lists **`validate-examples`** as a required
status check. A bare matrix would rename the checks to `validate-examples (minimal)`,
`validate-examples (westeros)`, … and the required `validate-examples` context would
never report — permanently blocking PRs. The aggregator keeps that exact check name
stable, green/red, regardless of how many entries the matrix expands to.

## Related issues

Closes #352.

## Review focus

- The dynamic matrix (`find … | jq -Rsc …`) and `fromJSON(needs.build-cli.outputs.examples)`.
- The aggregator's role in keeping the ruleset-required `validate-examples` check name stable.
- Workflow-injection safety: `matrix.example` is staged through `env:` before use in `run:`,
  per `.github/CLAUDE.md`.

## Testing

- `actionlint .github/workflows/validate-spec.yml` → exit 0.
- YAML parses; jobs are `build-cli`, `validate-example`, `validate-examples` (+ unchanged ones).
- Verified the discovery command locally against the real tree — yields all 8 examples:
  `["assertion-workflow","basic-family","complete-family","minimal","participant-assertions","single-file","temporal-properties","westeros"]`.
- Verified the per-leg command locally: `go build -o bin/glx ./glx` then
  `./bin/glx validate docs/examples/minimal/` → `✅ Archive is valid.`
- The workflow itself runs on this PR (CI is the real end-to-end check).

## Breaking changes

None. `validate-examples` remains a reported status check; only its internals change.


noahbjohnson (approved):
nice work 👍🏼

Issue #1055, opened 2026-06-08
Rewrite .github/CODEOWNERS: DRY both-maintainer catch-all, drop 4 redundant rules, honest header


isaacschepp:
## Problem

A comprehensive 2026 best-practices review of `.github/CODEOWNERS` (see also #258, #354, #421) found that the file is **non-enforcing today** — the "Main Protection" ruleset sets `require_code_owner_review: false`, so CODEOWNERS only *auto-requests* reviewers (enforcement is tracked separately in #258). Independent of that decision, the file content itself has two cosmetic-but-real defects:

1. **4 of 8 path rules change nothing** under CODEOWNERS' last-match-wins semantics (a matched owner set *replaces*, it does not merge with, earlier matches):
   - `/go-glx/`, `/glx/`, `/.github/` re-list the catch-all `* @noahbjohnson @isaacschepp` verbatim.
   - `/specification/schema/ @noahbjohnson` duplicates its `/specification/ @noahbjohnson` parent.

   These are no-ops and silent drift traps — when the catch-all owner set later changes (a third maintainer, a team handle), these lines keep the stale set on the repo's most important directories. The repo has hit this staleness failure mode before (#330 "commented out with stale paths"; the later stale-path fix in #600). GitHub's `codeowners/errors` API only catches *syntax*, never redundancy, so it will never flag these (`{"errors":[]}` today).

2. **Single-owner lines undercut bus-factor cross-review.** `/specification/`, `/docs/`, `/website/` auto-request only `@noahbjohnson`, leaving `@isaacschepp` off the spec/docs/website PRs. For a 2-person team the *primary* literature-backed value of CODEOWNERS is cross-review / knowledge transfer / bus-factor insurance (Bacchelli & Bird, *Modern Code Review*, ICSE 2013; Thongtanunam et al. 2016 on review-conferred ownership). Siloing the core deliverable (the spec) and the public docs to one person works against exactly that.

3. **Header overstates enforcement** — "individuals responsible for reviewing specific areas" implies a merge gate the enforcement layer does not currently deliver (the "protected-looking but unenforced" anti-pattern).

## Proposed Solution

Collapse to a DRY, both-maintainer catch-all that auto-requests **both** maintainers on every PR, drop all four redundant rules, and rewrite the header to be honest:

```
# GENEALOGIX Repository Code Owners
#
# Today this file ONLY AUTO-REQUESTS review from the listed owners -- it does
# NOT block merges, because the "Main Protection" ruleset sets
[...]

Issue #421, opened 2026-03-30
Add GOVERNANCE.md and MAINTAINERS.md for project transparency


isaacschepp:
## Summary

GLX has no formal governance documentation. As the project matures toward v1.0, contributors and potential adopters need to understand how decisions are made, who the maintainers are, and how to become one.

### Missing files

**GOVERNANCE.md** — Decision-making process:
- How spec changes are approved (majority? consensus? benevolent dictator?)
- Who has merge authority
- How disputes are resolved
- Voting mechanisms for breaking changes

**MAINTAINERS.md** — Current maintainer list:
- Names and GitHub handles
- Areas of responsibility
- Contact methods
- Emeritus maintainers

### Industry context

The CNCF project template (https://github.com/cncf/project-template) includes both files as standard for healthy open source projects. GitHub's 2026 open source report emphasizes that "clear pathways from contributor to reviewer to maintainer" are critical for scaling.

### Relationship to other issues

- #417 proposes a contributor ladder in CONTRIBUTING.md — this issue creates the dedicated files
- #413 proposes a formal RFC process — GOVERNANCE.md would document who approves RFCs
- CODEOWNERS (#330) is currently commented out — MAINTAINERS.md would be the human-readable equivalent

### Why this matters for a spec project

Potential adopters of GLX as a standard need to trust the governance. Standards bodies, genealogy software vendors, and institutional users evaluate governance transparency before committing to a format. GEDCOM's governance is opaque (invitation-only steering committee) — GLX can differentiate by being transparent.

isaacschepp:
## Update from a 2026 CODEOWNERS/governance review (2026-06-08)

This issue gains concrete content from the review, plus one correction.

### Stale premise corrected

The "Relationship to other issues" section says *"CODEOWNERS (#330) is currently commented out."* That's **no longer true** — #330 is fixed and `.github/CODEOWNERS` is active. The accurate framing: CODEOWNERS is active but **non-enforcing** (`require_code_owner_review: false` on the ruleset — see #258), and it is a glob-rules artifact, not a transparency document. MAINTAINERS.md remains the human-readable equivalent — that rationale stands, just for the updated reason.

### Concrete content for MAINTAINERS.md / GOVERNANCE.md

[...]

Issue #330, opened 2026-03-29
CODEOWNERS entirely commented out with stale paths


isaacschepp:
## Summary

`CODEOWNERS` has every single rule commented out. The file serves no purpose — no automatic review requests are being made. Additionally, the commented paths reference directories that don't exist or have moved.

## Current state

All 20+ rules are commented with `#`:
```
# * @genealogix/maintainers
# /specification/ @genealogix/specification-team
# /schema/ @genealogix/schema-team        ← stale: should be /specification/schema/
# /glx/ @genealogix/cli-developers
# /test-suite/ @genealogix/testing-team    ← stale: doesn't exist
# /examples/ @genealogix/example-maintainers  ← stale: should be /docs/examples/
```

Line 65 says: "If specific teams don't exist yet, these can be replaced with individual GitHub usernames"

### Option A: Activate with real usernames
Since the project has a small team, use individual usernames instead of teams:
```
* @noahbald
/specification/ @noahbald
/go-glx/ @noahbald
/glx/ @noahbald
/.github/ @noahbald
```

### Option B: Remove the file
If CODEOWNERS isn't needed yet, remove the file rather than keeping commented-out placeholder rules. It can be re-added when the team grows.

### Either way: fix stale paths
If kept, update the paths to match current structure:
- `/schema/` → `/specification/schema/`
- `/test-suite/` → remove (doesn't exist)
- `/examples/` → `/docs/examples/`
- `/specification/3-file-structure.md` → `/specification/3-archive-organization.md`

Issue #258, opened 2026-03-28
Enable branch protection: stale review dismissal, last-push approval, thread resolution


isaacschepp:
## Summary

The Main Protection ruleset has these disabled:
- `dismiss_stale_reviews_on_push: false` — allows merging after code changes without re-review
- `require_last_push_approval: false` — allows reviewer to approve their own force-push
- `required_review_thread_resolution: false` — allows merging with unresolved conversations

Also, CODEOWNERS is entirely commented out and `require_code_owner_review` is disabled.

## What to do

1. Enable all three settings in the branch ruleset
2. Uncomment at least one CODEOWNERS rule (e.g., `* @noahbjohnson @isaacschepp`)
3. Enable `require_code_owner_review`

isaacschepp:
## Update from a 2026 CODEOWNERS/governance review (2026-06-08)

Two corrections + two new findings that fall squarely in this issue's ruleset scope.

### Stale premise corrected

This issue's body says *"CODEOWNERS is entirely commented out"* — that is **no longer true**. #330 was fixed; `.github/CODEOWNERS` is now active with eight rules, and the `codeowners/errors` API returns `{"errors":[]}`. So step 2 ("uncomment a rule") is already done. What remains here is the **ruleset** work.

### Live ruleset state (`Main Protection`, id `10185426`, active, targets `~DEFAULT_BRANCH`)

```
required_approving_review_count: 1
require_code_owner_review:        false      <-- CODEOWNERS gates nothing
dismiss_stale_reviews_on_push:    false
require_last_push_approval:       false
required_review_thread_resolution: false
allowed_merge_methods:            [squash]
bypass_actors:                    [{actor_type: OrganizationAdmin, bypass_mode: pull_request}]
```

### New finding 1 — CODEOWNERS is decorative (HIGH)

Because `require_code_owner_review: false`, CODEOWNERS only **auto-requests** reviewers; it does not block merges. Every per-path ownership rule is advisory. Any PR merges with 1 approval from anyone with write access, regardless of CODEOWNERS. This is the single fact that reframes the whole CODEOWNERS discussion (see the new CODEOWNERS-rewrite issue, which is non-enforcing cleanup; the *enforcement* decision lives here).

### New finding 2 — admin bypass of the 1-approval rule (MEDIUM)

[...]

PR #1088, opened 2026-06-10
ci: validate CODEOWNERS on every change

### LINKED ISSUE #354: CI: add CODEOWNERS validation to prevent stale entries
## Problem

Once CODEOWNERS is activated (#330), there is no CI check to prevent it from going stale again. Stale entries (renamed paths, departed users, deleted teams) can silently break reviewer assignment or — worse — block PR merges entirely if enforcement is enabled.

GitHub's web UI highlights syntax errors when viewing the file, but this requires someone to manually check. The REST API (`GET /repos/{owner}/{repo}/codeowners/errors`) exists but has a known bug where it sometimes returns empty errors in private repos.

## Proposed Solution

Add a CI job that validates CODEOWNERS on every PR that modifies it:

```yaml
name: Validate CODEOWNERS

on:
  pull_request:
    paths:
      - '.github/CODEOWNERS'
      - 'CODEOWNERS'
      - 'docs/CODEOWNERS'

permissions:
  contents: read

jobs:
  validate:
    runs-on: ubuntu-latest
    timeout-minutes: 5
    steps:
      - uses: actions/checkout@v6

      - name: Validate CODEOWNERS
        uses: mszostok/codeowners-validator@v0.7.4
        with:
          checks: 'syntax,owners,duppatterns,files'
          experimental_checks: 'avoid-shadowing'
          github_access_token: ${{ secrets.GITHUB_TOKEN }}
```

### What this validates

| Check | What it catches |
|-------|----------------|
|  | Malformed patterns, invalid team references |
|  | Users/teams that don't exist or lack write access |
|  | Duplicate path patterns (only last one takes effect) |
|  | Paths in CODEOWNERS that don't match any files in the repo |
|  (experimental) | Rules that are completely overridden by later rules and never take effect |

### Tool Options

| Tool | Language | Key advantage |
|------|----------|---------------|
| [mszostok/codeowners-validator](https://github.com/mszostok/codeowners-validator) | Go | Most mature, native GitHub Action, shadow detection |
| [dotanuki-labs/canopus](https://github.com/dotanuki-labs/canopus) | Rust | Fastest, has repair functionality, active in 2026 |
| [github-community-projects/cleanowners](https://github.com/github-community-projects/cleanowners) | Python | Built by GitHub OSPO, auto-removes departed org members |

### Complementary: scheduled cleanowners audit

For ongoing maintenance, run  weekly to detect org members who have left:

```yaml
name: Audit CODEOWNERS

on:
  schedule:
    - cron: '0 6 * * 1'  # Weekly Monday 06:00 UTC

jobs:
  audit:
    runs-on: ubuntu-latest
    steps:
      - uses: github/cleanowners@v1
        with:
          dry_run: true  # Report only, don't auto-fix
```

## Why

- Prevents regression to a stale/broken CODEOWNERS after #330 is fixed
- Shadow detection catches "last match wins" ordering mistakes — the single most common CODEOWNERS bug
- File existence check catches path drift when directories are renamed/moved
- Owner validation catches team/user references that don't have write access (which silently blocks merges if enforcement is on)
- The 3 MB file size limit is silently enforced — validation can warn before that threshold

## Dependencies

- Depends on #330 (activate CODEOWNERS first)

isaacschepp:
Implements #354 following the refinements in [the 2026-06-08 review comment](https://github.com/genealogix/glx/issues/354#issuecomment-4651828763) (#330 is closed, so this was unblocked).

## What's in here

**`.github/workflows/codeowners.yml`** — runs on any PR/push touching a CODEOWNERS file (all three GitHub-recognized locations), the workflow itself, or the `ci-tools` tool pins:

1. **`mszostok/codeowners-validator`** with `checks: files,duppatterns,syntax` + experimental `avoid-shadowing` — all purely local, no token. SHA-pinned to `7f3f5e2` = `v0.7.4`; verified via `gh api repos/mszostok/codeowners-validator/tags` that upstream publishes only `vX.Y.Z` tags (no floating `@v0.7`), with an inline comment so nobody "tidies" the pin, per `.github/CLAUDE.md`. Also verified `avoid-shadowing` exists at v0.7.4 in `internal/load/load.go` (the action.yml input description listing only `notowned` is stale upstream).
2. **Zero-secret owner-existence floor** — `gh api users/<handle>` for every non-comment `@owner`. This is the renamed/departed-handle catch and needs no special scope.
3. **Full validator `owners` check (write access), gated on an optional secret.** The validator's `CheckSatisfied` preflight hard-requires the `read:org` classic scope for plain tokens, and Actions' `GITHUB_TOKEN` is an installation token with no classic scopes — the first run of this PR failed with `missing scopes: "read:org"`, exactly as the review comment anticipated ("optional, needs `read:org`"). The step runs only when a `CODEOWNERS_VALIDATOR_TOKEN` secret (classic PAT with `read:org`) is configured; absent the secret it is skipped, not failed.

**`make lint-codeowners`** — the zero-token floor from the review comment: `hmarr/codeowners --unowned`, pinned at `v1.2.1` via the `ci-tools/go.mod` tool directive (same pattern as `govulncheck`). The CLI always exits 0, so the target fails explicitly on non-empty output. Wired into `make check` and run as a final CI step.

## Deliberate divergences from the issue/comment

- **`notowned` + `NOT_OWNED_CHECKER_SKIP_PATTERNS=*` is omitted.** That suggestion predates #1055: with today's catch-all-only file, skipping `*` ignores the only rule and flags the entire tree as unowned. Coverage is handled by `lint-codeowners` instead, which also guards against the catch-all ever being deleted.
- **Scheduled `cleanowners` audit is deferred.** It detects departed *org members in teams*; with two individual owners, the existence floor on every CODEOWNERS change covers the same ground. Revisit if a team owner is ever added.

## Documented caveat (from the review comment's AC note)

No tool here catches a mistyped-but-existing username or a functionally-redundant same-owner rule — documented in the workflow header. Auto-request behavior should be confirmed on a real PR after any future owner-set change; this PR itself exercises the workflow (it modifies `.github/workflows/codeowners.yml`, which is in its own path filter).

## Verification

- `make lint-codeowners` passes locally against the current tree, and correctly fails (with the unowned file list) when pointed at a CODEOWNERS lacking the catch-all.
- Owner-extraction snippet tested against the real file: comment-line `@mentions` are stripped, yields exactly `@isaacschepp` + `@noahbjohnson`.
- `go -C ci-tools mod tidy -diff` clean.
- The workflow runs on this PR — see checks below.

Related: #258 (enforcement flag), #421 (governance docs), #1055 (file rewrite).

Closes #354

🤖 Generated with [Claude Code](https://claude.com/claude-code)


isaacschepp:
@copilot resolve the merge conflicts in this pull request

noahbjohnson (approved):

PR #1097, opened 2026-06-11
ci: Make the test-conformance warning-gate synchronous and broaden its match

### LINKED ISSUE #1033: validate-spec.yml: test-conformance Warning-gate races an async process-substitution `tee` (non-deterministic false-pass)
## Problem

The `test-conformance` job's "fail on `Warning:` stderr" gate is non-deterministic and can silently pass a run that actually emitted warnings. It also misses non-line-anchored warnings.

## Evidence (`.github/workflows/validate-spec.yml:194-198`)

```yaml
      - name: Run Go Tests
        run: |
          ...
          go test ./glx/... ./go-glx/... ./tools/... -v -timeout 15m -coverprofile=coverage.out 2> >(tee /tmp/test-stderr.log >&2)
          if grep -qi "^Warning:" /tmp/test-stderr.log 2>/dev/null; then
            echo "::error::Tests emitted warnings on stderr — see log above"
            exit 1
          fi
```

Two defects:

1. **Async race (primary).** Commands inside `>( ... )` run asynchronously; bash issues no `wait()` for the process-substitution child before executing the next command. The GNU Bash manual states under *Process Substitution*: "The process list is run asynchronously." So the `grep` can open `/tmp/test-stderr.log` before `tee` has flushed/closed it, missing warnings written near `go test`'s exit. The miss rate is small and load-dependent (the signature of a genuine async race) — it will pass ~99% of the time and flap rarely, which is the worst failure mode for a gate. The `2>/dev/null` on the grep additionally masks a "file not yet created" error.
2. **Regex narrowness (secondary, deterministic).** `^Warning:` matches only warnings printed at the very start of a line; an indented or mid-line `Warning:` (e.g. from a test logger) is never caught even when timing is perfect.

## Why it matters

The warning-gate is the *entire reason* these lines exist. A genuine `go test` failure still fails the step via `set -e`, so the blast radius is bounded to "a stderr warning slips past unnoticed" — but that is precisely the signal this step is meant to enforce, and it does so unreliably. ~20+ `glx/*_runner.go` call sites emit `Warning:` to stderr and are exercised by tests, so a test leaking an unexpected warning is a realistic scenario the gate should catch.

## Proposed fix

Use a synchronous pipeline so the file is complete before grep runs, and broaden the regex:

```yaml
      - name: Run Go Tests
        run: |
          set -o pipefail
          go test ./glx/... ./go-glx/... ./tools/... -v -timeout 15m \
            -coverprofile=coverage.out 2>&1 | tee /tmp/test-output.log
          if grep -qiE '(^|[[:space:]])Warning:' /tmp/test-output.log; then
            echo "::error::Tests emitted warnings — see log above"
            exit 1
          fi
```

The `| tee` form is fully synchronized (the next command runs only after `tee` exits). If stderr must stay separate from stdout, use a plain synchronous redirect (`2> /tmp/test-stderr.log`, closed when `go test` exits) and capture `go test`'s exit code explicitly to preserve pipefail semantics — do **not** rely on adding `wait` cosmetically. Note merging stderr into stdout (`2>&1`) changes what is matched; pick deliberately.

## 2026 references

- GNU Bash Reference Manual, *Process Substitution*: "The process list is run asynchronously" — <https://www.gnu.org/software/bash/manual/bash.html#Process-Substitution>
- Nate Woodward, "Process Substitution And Race Conditions" (2019-11-25): output files "will not be available in time for processing; files will be empty while commands are running" — <https://natewoodward.org/blog/>
- comp.unix.shell, "Race condition with pipe expression in process substitution (ksh, bash, but not zsh)" — <https://groups.google.com/g/comp.unix.shell>
- Greg's Wiki, ProcessSubstitution (the bash `wait "$!"` mechanism exists because synchronization is otherwise absent) — <https://mywiki.wooledge.org/ProcessSubstitution>

## Related issues

Distinct from #347 (Go `-race` data-race detector — a different layer) and #351 (step summaries). No existing issue covers this bash-level race. Introduced in #1018 (commit 3fa9b734).

isaacschepp:
## What and why

The `test-conformance` job's "fail on `Warning:` stderr" gate was non-deterministic. It captured `go test` stderr through an async process-substitution `tee` (`2> >(tee /tmp/test-stderr.log >&2)`); bash runs that process list asynchronously and issues no `wait()` for it, so the following `grep` could open the log before `tee` had flushed/closed it — silently passing a run that actually emitted warnings near process exit. A rare, load-dependent false-pass is the worst failure mode for a gate. The `^Warning:` anchor was a second, deterministic gap: indented or mid-line warnings were never matched.

The fix:

- **Synchronous capture.** Stderr goes through a plain redirect (`2> /tmp/test-stderr.log`), which is complete by the time the next command runs. The log is then surfaced with `cat` so warnings remain visible in the job output.
- **Explicit exit-code propagation.** `|| status=$?` + `exit "$status"` preserves a genuine `go test` failure (previously handled by `set -e`) while still printing the captured stderr first — strictly better than before, where a test failure aborted the step ahead of the gate.
- **Broader match.** `grep -qiE '(^|[[:space:]])Warning:'` catches indented and mid-line warnings (e.g. from a test logger), per the regex proposed in the issue.
- **Deliberately not `2>&1`.** The issue offered a stdout-merging `| tee` variant as an alternative; merging would let `go test -v` stdout that legitimately contains the literal string (golden output, `t.Log` of expected warnings — `Warning:` appears widely in testdata and test expectations) trip the gate with false failures. Keeping stderr separate preserves the gate's original scope exactly.

The trade-off vs. the old form: stderr is no longer streamed live during the test run — it appears in the log when the run finishes. For a 15-minute CI step whose output is read after the fact, determinism is worth far more than live streaming.

## Related issues

Fixes #1033

## Review focus

The bash in the `Run Go Tests` step — specifically that the exit-code handling is correct under GitHub's default `bash -e` shell, and the choice to keep stderr separate rather than merge with stdout.

## Testing

Simulated the new step logic locally against six cases: line-anchored, indented, and mid-line warnings (all caught → gate fails), clean stderr (passes), a failing `go test` stand-in (step fails with the original exit code after stderr is surfaced), and a `forewarning:`-style near-miss (no false positive — the regex requires start-of-line or whitespace before `Warning:`).

## Breaking changes

None.


noahbjohnson (dismissed):

isaacschepp:
@copilot resolve the merge conflicts in this pull request

noahbjohnson (approved):

Issue #1136, opened 2026-07-27
glx init should create .gitkeep files so the scaffold survives a clone


noahbjohnson:
Follow-up from review feedback on #1084 ([comment](https://github.com/genealogix/glx/pull/1084#issuecomment-4937958773)), where this was explicitly deferred as out of scope for that docs PR.

### Area

cli

### Current limitation

`glx init` scaffolds eleven entity-type directories but leaves them empty. Git does not track empty directories, so none of them survive a commit — only `vocabularies/` (which has content) and the two root files do.

Reproduced against `main`:

```console
$ glx init archive && cd archive && git init -q . && git add . && git commit -qm init
$ git ls-files | grep -c vocabularies/    # 25
$ git ls-files | grep -cE '^(persons|events|sources)/'    # 0

$ cd .. && git clone -q archive clone && find clone -type d -not -path '*/.git*'
clone
clone/vocabularies
```

Two consequences:

**1. A documented command fails for anyone who clones the archive.** The Research Cycle section added in #1084 tells readers to stage entity directories by name. On the author's own machine those directories exist, so it works; on a fresh clone it does not:

```console
$ git add persons/ events/ citations/ sources/
fatal: pathspec 'persons/' did not match any files
$ echo $?
128
```

This is the exact clone-and-collaborate workflow the guide's Remote Backup and Collaboration sections encourage, so the failure lands on the collaborator rather than the author. #1084 addresses the docs side by switching that line to `git add -A`; this issue is about the underlying scaffold.

**2. The scaffold stops being self-documenting.** The point of creating eleven empty directories is to show a new user where each entity type goes. A collaborator who clones sees only `vocabularies/`, losing that affordance precisely when they are least familiar with the layout.

### Proposed improvement

Have `createMultiFileArchive` (`glx/init_runner.go:96`) write a `.gitkeep` into each entity-type directory it creates, alongside the existing `.gitignore` and `README.md`. The directory list is already derived from `glxlib.AllEntityTypes`, so new entity types would be covered automatically.

Worth deciding during implementation:

- **Removal on first use.** Whether `.gitkeep` should be deleted once a directory holds real entities, or left permanently. Leaving it is simpler and harmless; removing it keeps archives tidy but adds a cleanup path to every entity write.
- **Naming.** `.gitkeep` is convention, not a Git feature — any tracked file works. It does tie a general-purpose scaffold to Git specifically, though the project is explicitly Git-native, so that seems acceptable.
[...]

PR #1138, opened 2026-07-27
docs: Fix three broken copy-paste snippets in Git Workflow Guide

### LINKED ISSUE #1137: docs: Three broken copy-paste snippets in the Git Workflow Guide
#1084 merged (629f96b) without the three snippet fixes raised in [review](https://github.com/genealogix/glx/pull/1084#issuecomment-4937958773), so they are now live in the published guide. Capturing them here.

All three are copy-paste failures: each block is syntactically fine and looks right on the page, but breaks when a reader actually runs it. That matters more than usual for this guide, since its value is that people copy these blocks verbatim. Re-verified against `origin/main` today, not just restated from the review.

---

### 1. The CI workflow's checksum step fails on every run

`docs/guides/git-workflow.md:221-224`. The release asset is downloaded under a new name, but the verification line asks `sha256sum` to check the *original* name, which is not on disk:

```yaml
curl -fsSL -o glx.tar.gz "$base/glx_Linux_x86_64.tar.gz"
curl -fsSL -o checksums.txt "$base/checksums.txt"
grep ' glx_Linux_x86_64.tar.gz$' checksums.txt | sha256sum -c -
tar -xzf glx.tar.gz -C .bin glx
```

Reproduced:

```console
$ grep ' glx_Linux_x86_64.tar.gz$' checksums.txt | sha256sum -c -
sha256sum: glx_Linux_x86_64.tar.gz: No such file or directory
glx_Linux_x86_64.tar.gz: FAILED open or read
$ echo $?
1
```

GitHub Actions runs `run:` steps under `bash -eo pipefail`, so that non-zero exit fails the job. Anyone who copies this workflow gets a permanently red validation gate regardless of version or network — and the failure mode (a checksum step that appears to reject the release) invites exactly the wrong diagnosis.

**Fix** — stop renaming; download and extract under the checksum's own name:

```yaml
curl -fsSL -o glx_Linux_x86_64.tar.gz "$base/glx_Linux_x86_64.tar.gz"
curl -fsSL -o checksums.txt "$base/checksums.txt"
grep ' glx_Linux_x86_64.tar.gz$' checksums.txt | sha256sum -c -
tar -xzf glx_Linux_x86_64.tar.gz -C .bin glx
```

### 2. The merge-conflict Person snippet is invalid GLX

`docs/guides/git-workflow.md:249-261`, the "When They Happen Anyway" example, puts a top-level `names:` list as a sibling of `properties:`. The Person schema is `additionalProperties: false` and permits only `properties` and `notes`:

```console
$ glx validate t.glx
Found 1 structural errors in 1 files:
- Error in t.glx: /persons/person-john-smith: additional properties 'names' not allowed
structural validation failed
```

Doubly unfortunate because the same guide wires `glx validate` into a pre-commit hook a few sections earlier — a reader who resolves the conflict exactly as shown then cannot commit the result.

**Fix** — reshape so the name lives under `properties.name` (`value` / `fields.given` / `fields.surname`), matching the real examples elsewhere in the docs, with the conflict markers still around `occupation` so the teaching point survives.

### 3. The Research Cycle's `git add` fails on a fresh clone

`docs/guides/git-workflow.md:63` stages entity directories by name:

```console
$ git add persons/ events/ citations/ sources/
fatal: pathspec 'persons/' did not match any files
$ echo $?
128
```

`glx init` creates those directories empty, Git does not track empty directories, so a clone does not have them — only `vocabularies/` (which has content) survives. It works on the author's own machine and fails for the collaborator, which is the harder case to spot in review.

**Fix** — `git add -A` (or `git add .`, matching the Setting Up section at line 23).

The underlying scaffold behavior is tracked separately in #1136; that is a `glx init` change, whereas this one is a one-line docs fix that stands on its own.

---

### Not affected

Everything else in the guide was exercised and came back clean: the `.gitattributes` block, default-archive validation, `glx diff --short`, the CHANGELOG entry, and the nav/sidebar wiring. The conceptual content — the `git merge` vs `glx merge` distinction, the "why conflicts are rare" reasoning, and all links and anchors — was checked against the code and holds up. This issue is strictly about the three runnable snippets.

noahbjohnson:
## What and why

Fixes the three broken copy-paste snippets in `docs/guides/git-workflow.md` that #1084 merged with. Each block is syntactically valid and looks right on the page, but fails when a reader actually runs it — which matters more than usual here, since the guide's value is that people copy these blocks verbatim.

**1. The CI workflow's checksum step failed on every run.** The release asset was downloaded as `glx.tar.gz`, but `sha256sum -c` was asked to verify `glx_Linux_x86_64.tar.gz`, which was no longer on disk. A `run:` step with no explicit `shell:` key runs under GitHub Actions' default `bash -e {0}`, so the exit 1 failed the job regardless of version or network — and the failure presents as a rejected checksum, inviting exactly the wrong diagnosis. The asset now keeps its published name through download, verification, and extraction.

**2. The merge-conflict Person snippet was invalid GLX.** A top-level `names:` list sat beside `properties:`; the Person schema is `additionalProperties: false`. Doubly awkward because the guide wires `glx validate` into a pre-commit hook a few sections earlier, so a reader who resolved the conflict exactly as shown then couldn't commit the result. The name now sits under `properties.name` with `value`/`fields`, matching the shape used in the Quickstart — conflict markers stay around `occupation` so the teaching point is unchanged.

**3. The Research Cycle's `git add` failed on a fresh clone.** `git add persons/ events/ citations/ sources/` exits 128 with `fatal: pathspec 'persons/' did not match any files`, because `glx init` creates those directories empty and Git doesn't track empty directories. It works on the author's machine and fails for the collaborator — the exact clone-and-collaborate workflow the guide's Collaboration section encourages. Now `git add -A`, matching the Setting Up section. The underlying scaffold behavior is a `glx init` change tracked separately in #1136; this is the docs-side fix that stands on its own. Review feedback added a following sentence noting that `git add -A` stages the whole archive, so a reader with unrelated work in progress should stage paths explicitly.

## Related issues

Fixes #1137

Follow-up from review feedback on #1084, which merged before these were addressed.

Review of this PR also flagged that the CI snippet's checksum step verifies integrity but not authenticity (`checksums.txt` arrives over the same unauthenticated channel). That is out of scope here and is filed as #1173.

## Review focus

The reshaped Person snippet in "When They Happen Anyway" — whether `properties.name` with `value`/`fields.given`/`fields.surname` is the form you want shown here. I matched `docs/quickstart.md:69-77`, which uses the same `person-john-smith` ID, minus the `type: "birth"` field and `sex` (not relevant to the conflict being illustrated).

Otherwise trivial: two one-line command changes plus one added sentence of prose.

## Testing

Each fix was reproduced as broken against merged `main`, then re-verified after the change.

1. Checksum step — before: `sha256sum: glx_Linux_x86_64.tar.gz: No such file or directory` → `FAILED open or read`, exit 1. After: `glx_Linux_x86_64.tar.gz: OK`, exit 0.
2. Person snippet — before: `glx validate` → `additional properties 'names' not allowed`, structural validation failed. After (markers resolved): passes structural and semantic validation, `occupation` included.
3. `git add` — before, on a real `git clone` of a `glx init` archive: `fatal: pathspec 'persons/' did not match any files`, exit 128. After: `git add -A` exits 0 and stages the entity file.

Also run:

- `npm run build` (VitePress) — completes, no dead links
- `npx markdownlint-cli2` on both changed files — clean (the one repo-wide finding is a pre-existing MD040 in `docs/superpowers/specs/`, untouched here)
- `make check-links` is Linux-only on this machine (the script needs GNU awk's 3-arg `match()`; BSD awk errors out), so it's left to CI. No links were changed.

## Breaking changes

None — documentation only.


CHANGELOG.md
```diff
@@ -79,6 +79,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
...permanently red validation gate that looked like a rejected checksum; the asset now keeps its published name through download, verification, and extraction. (2) The merge-conflict example put a top-level `names:` list beside `properties:`, which `glx validate` rejects (`additional properties 'names' not allowed`, the Person schema being `additionalProperties: false`) — doubly awkward because the same guide wires `glx validate` into a pre-commit hook a few sections earlier, so a reader who resolved the conflict as shown then could not commit; the name now sits under `properties.name` with `value`/`fields`, matching the Quickstart, with the conflict markers still around `occupation`. (3) The Research Cycle staged entity directories by name (`git add persons/ events/ …`), which fails with `fatal: pathspec 'persons/' did not match any files` (exit 128) on a fresh clone, because `glx init` creates those directories empty and Git does not track empty directories — it worked on the author's machine and failed for the collaborator, the exact workflow the guide's Collaboration section encourages; now `git add -A`. The underlying scaffold behavior is tracked separately in #1136. Fixes #1137.
```
isaacschepp, line 82: The shell attribution here is off. A `run:` step with no explicit `shell:` key — which is what the guide's snippet is — runs under GitHub Actions' default `bash -e {0}`; `pipefail` is only added when the step sets `shell: bash` explicitly (`bash --noprofile --norc -eo pipefail {0}`).

The conclusion is unaffected: `sha256sum` is the last command in the pipeline, so its exit 1 *is* the pipeline's status and `-e` fails the step either way. But this entry ships as documentation on the rendered changelog page, and a reader who takes the `pipefail` claim as a statement about default `run:` behavior will carry a wrong mental model into their own workflows.

Suggest dropping the flag name, e.g. "…the resulting exit 1 failed the job on every run under GitHub Actions' error-on-exit default". (The same phrasing appears in the PR description — worth fixing there too, though only the CHANGELOG actually ships.)

docs/guides/git-workflow.md
```diff
@@ -60,7 +60,7 @@ Always validate before committing, so every commit in history is a known-good ar
 
 ```bash
 glx validate
-git add persons/ events/ citations/ sources/
+git add -A
```
isaacschepp, line 63: Correct fix — `git add persons/` really does exit 128 on a fresh clone, and I confirmed no other doc in `docs/` or `specification/` stages entity directories this way except the deliberate commit-by-entity-type example at `specification/3-archive-organization.md:446`.

Optional: `-A` stages the *entire* working tree, which sits a little awkwardly against this section's own "one source examined, one commit" advice — a researcher with unrelated edits in flight will sweep them into the census commit. A half-sentence after the block would keep both the working command and the teaching point, e.g. "`git add -A` stages everything in the archive; if you have unrelated work in progress, stage those paths explicitly instead."

Take it or leave it — the command as written is correct and copy-pasteable, which is the point of the PR.

docs/guides/git-workflow.md
```diff
@@ -218,10 +218,10 @@ jobs:
           # replace "latest/download" with a pinned "download/vX.Y.Z".
           mkdir -p .bin
           base="https://github.com/genealogix/glx/releases/latest/download"
-          curl -fsSL -o glx.tar.gz "$base/glx_Linux_x86_64.tar.gz"
+          curl -fsSL -o glx_Linux_x86_64.tar.gz "$base/glx_Linux_x86_64.tar.gz"
           curl -fsSL -o checksums.txt "$base/checksums.txt"
           grep ' glx_Linux_x86_64.tar.gz$' checksums.txt | sha256sum -c -
-          tar -xzf glx.tar.gz -C .bin glx
+          tar -xzf glx_Linux_x86_64.tar.gz -C .bin glx
```
isaacschepp, line 224: Not a change request for this PR — flagging as a follow-up, since this PR is what makes the block actually run.

The block now verifies the tarball against `checksums.txt`, but `checksums.txt` itself arrives over the same unauthenticated channel as the tarball. That catches corruption and truncated downloads; it does not catch a substituted release, which is what the comment's "never pipe an unverified download straight into a shell" is warning about.

The repo already publishes what closes this: `.goreleaser.yml`'s `signs:` block runs `cosign sign-blob` over `artifacts: checksum`, producing `checksums.txt.sigstore.json`. A `cosign verify-blob --bundle checksums.txt.sigstore.json --certificate-identity … --certificate-oidc-issuer https://token.actions.githubusercontent.com checksums.txt` step ahead of the `sha256sum` line would make the guide's security claim real, at the cost of an extra install step in the example. Happy to file the issue if you want it tracked separately.

isaacschepp (changes requested):
## Overview

Three independent runtime fixes to `docs/guides/git-workflow.md`, plus a matching `## [Unreleased] → Fixed` CHANGELOG entry. I verified each against the repo rather than taking the description on faith:

1. **Checksum step** — `.goreleaser.yml`'s `name_template` (`{{ .ProjectName }}_{{ title .Os }}_x86_64` for `amd64`) does resolve to `glx_Linux_x86_64.tar.gz`, and `checksum.name_template` is `checksums.txt`. Downloading under the published name makes `grep ' glx_Linux_x86_64.tar.gz$' checksums.txt | sha256sum -c -` resolve against a file that exists, and `tar -xzf … -C .bin glx` still extracts the root-level binary (`archives.files` only adds LICENSE/NOTICE/README/CHANGELOG alongside it). Correct.
2. **Person snippet** — `specification/schema/v1/person.schema.json` is `"additionalProperties": false` with only `properties` and `notes` allowed, so the old top-level `names:` was genuinely rejected. The new shape matches `docs/quickstart.md:69-77` exactly (minus `type`/`sex`), and both `name` and `occupation` are registered in `specification/5-standard-vocabularies/person-properties.glx`. Correct, and the right form to show here — no reservations on the "Review focus" question.
3. **`git add -A`** — `git add persons/` on a fresh clone does fail with exit 128 when `glx init`'s empty directories were never tracked. Correct.

I also checked the fixes are complete rather than local: `git grep` finds no other `- given:` / top-level `names:` person examples anywhere in `docs/` or `specification/`, and no other release-asset download block. The one remaining `git add persons/` (`specification/3-archive-organization.md:446`) is a deliberate commit-by-entity-type example that runs after files exist, so it is not the same defect.

All 30+ checks pass; branch is mergeable; no prior reviews on the PR.

## Requested change

One factual correction in the CHANGELOG (inline). Everything else below is optional.

## Non-blocking suggestions

- **`git add -A` vs the "one source examined, one commit" discipline** the same section teaches — see inline note. Also, the description says this matches the Setting Up section, which actually uses `git add .`; `-A` is the better choice for a guide (it is repo-root-scoped regardless of cwd), so I would keep `-A` and not chase the wording.
- **The checksum block verifies integrity, not authenticity** — worth a follow-up issue rather than a change here; see inline.

## Risks

None. Documentation-only, no schema or code surface touched, and the reproduce-then-verify testing in the description is consistent with what I found in the repo.

noahbjohnson:
**This merge was unintentional and has been unwound.** It is recorded as merged, but its commit is no longer reachable from `main` or any open branch.

This PR had auto-merge (squash) armed since 2026-07-27. The gates holding it come from the `Main Protection` ruleset, which targets `~DEFAULT_BRANCH` only — so when the base was briefly repointed to `chore/consolidated-dep-updates` (an unprotected branch) at 06:56:43Z, they stopped applying and the queued auto-merge fired 4 seconds later.

The merge commit has been removed from `chore/consolidated-dep-updates`, which is back to dependency changes only (#1172).

**The work is not lost.** It is recreated verbatim — including the review follow-up correcting the `pipefail` attribution in the CHANGELOG and the added note about `git add -A` scope — as **#1178**, based on `main`. Please review there.

Issue #1141, opened 2026-07-27
go-glx: library-level GEDZIP import over fs.FS


noahbjohnson:
## Problem

GEDZIP import is CLI-only: `glx/gedzip_import.go` (package main) extracts the archive to a temp dir and delegates to the file-path-based `importGEDCOM`. Library consumers must reimplement the entire zip layer themselves — including the security-sensitive parts (name validation, duplicate detection, traversal rejection).

## Proposal

```go
// go-glx
func ImportGEDZIP(bundle fs.FS, logW io.Writer) (*GLXFile, *ImportResult, error)
```

Mirrors the existing `ImportGEDCOM(reader io.Reader, logWriter io.Writer)` shape (`go-glx/gedcom_import.go:351`).

The no-I/O stance is preserved by `fs.FS`: `zip.Reader` implements it, members stream on open, nothing is buffered whole, and the **caller owns storage** (`zip.Reader`, `os.DirFS`, `fstest.MapFS`). Tests stay filesystem-free via `fstest.MapFS`. `go-glx/CLAUDE.md` should be updated to explicitly bless `fs.FS` alongside `io.Reader`/`io.Writer`.

## Requirements

1. **Root GEDCOM discovery.** The GEDZIP spec mandates a root entry literally named `gedcom.ged`, and the current CLI enforces exactly that (`gedzipGedcomEntry`, `hasGedcomEntry`). This proposal deliberately loosens it for real-world vendor exports: prefer `gedcom.ged`; otherwise fall back to case-insensitive `.ged`/`.gedcom` discovery (exactly one required — multiple is an error), tolerating a single wrapper directory. The fallback emits a warning on `ImportResult`. **Note: via requirement 5 this loosens what the CLI accepts today** — `ErrGEDZIPMissingGedcom` semantics and error text change accordingly.
2. **Media resolution.** Resolve each `MediaFiles[].RelativePath` against the FS, relative to the `.ged`'s directory: reject raw `..` segments and absolute paths **before** `path.Clean`; normalize `\` → `/` (inherited from #1139); exact match first, then case-insensitive; no basename guessing. The GEDCOM 7.0 percent-decoding fallback (currently in `copyMediaFile`, `glx/media_copy.go:128-137`) must survive the move — 7.0 FILE refs are percent-encoded and dropping it would break resolution for compliant files. Unresolved refs → warning on `ImportResult`, not an error.
[...]

PR #1142, opened 2026-07-27
feat: Export GEDCOM 5.5.1 BLOB decoder as go-glx DecodeGEDCOMBlob

### LINKED ISSUE #1140: go-glx: export GEDCOM 5.5.1 BLOB decoding (move decodeGEDCOMBlob from CLI into the library)
## Problem

`MediaFileSource.BlobData` (`go-glx/gedcom_import.go:80-82`) carries raw GEDCOM 5.5.1 BLOB text, but the decoder lives in the CLI: `decodeGEDCOMBlob` in `glx/media_copy.go:187-249` (package main). Library consumers who receive `BlobData` must reimplement the 5.5.1 decoding themselves — including the validation and trailing-group edge cases the CLI version already handles.

## Proposal

This is a **move, not new work**: relocate the existing decoder into go-glx and export it.

```go
// go-glx
func DecodeGEDCOMBlob(blobText string) ([]byte, error)
```

- Pure function, no I/O — fully compliant with the library's no-filesystem rule.
- The three sentinel errors move with it into the library's public API: `ErrEmptyBlobData`, `ErrInvalidBlobLength`, `ErrInvalidBlobChar` (currently `glx/errors.go:24-26`).
- Name it `DecodeGEDCOMBlob`, not `DecodeBlob` — the package is `glx`, so `glx.DecodeBlob` would read as "decode a GLX blob".
- CLI's `copyMediaFiles` becomes a caller; existing tests move to go-glx alongside.

Prerequisite for a future library-level GEDZIP import (which needs to decode BLOB entries without depending on the CLI).

noahbjohnson:
## What and why

Moves the GEDCOM 5.5.1 BLOB decoder from the CLI into the library and exports it. `MediaFileSource.BlobData` carries raw BLOB text, but the decoder lived unexported in `glx/media_copy.go` (package main), so library consumers had to reimplement the 5.5.1 decoding — including validation and the trailing-group edge cases — themselves.

- New `go-glx/gedcom_blob.go`: exported `DecodeGEDCOMBlob(blobText string) ([]byte, error)` plus three sentinel errors (`ErrGEDCOMBlobEmpty`, `ErrGEDCOMBlobLength`, `ErrGEDCOMBlobChar`), moved out of `glx/errors.go` where they were named `ErrEmptyBlobData`, `ErrInvalidBlobLength`, and `ErrInvalidBlobChar`. Pure function, no I/O — compliant with the library's no-filesystem rule.
- Named `DecodeGEDCOMBlob` rather than `DecodeBlob`: the package is `glx`, so `glx.DecodeBlob` would read as "decode a GLX blob". The sentinels carry the same `GEDCOMBlob` prefix for the same reason, matching the in-package precedent set by `go-glx/census.go`.
- `copyMediaFiles` in the CLI becomes a caller; the CLI-local decoder and tests are removed.

Prerequisite for library-level GEDZIP import (#1141), which needs to decode BLOB entries without depending on the CLI.

### Behavior changes made while exporting

Four deliberate changes, all made now because this becomes public API on merge:

1. **Whitespace tolerance widened.** `blobWhitespace` now strips `\t`, `\v`, and `\f` alongside space, `\r`, and `\n`. All three are below `'.'` (0x2E), so a BLOB padded with them previously failed with the invalid-character error and now decodes. This is strictly widening: no input that decoded before decodes differently, and nothing that decoded before now fails.
2. **Invalid-character errors are uniformly wrappable.** The full-group path now wraps `ErrGEDCOMBlobChar` like the trailing-group path already did, so every invalid-character failure is detectable via `errors.Is`. The message gains the sentinel's "must be in range '.' to 'm'" hint.
3. **The offending byte is reported as hex as well as a quoted character.** `%q` formats a `byte` as a rune, so on its own it renders any byte >= 0x80 as the Latin-1 character with that code point rather than the byte actually in the input (0xFF as `'ÿ'`). The message now carries `%q (0x%02X)`; the hex cannot misreport.
4. **Invalid characters take precedence over invalid length.** Trailing-group characters are validated before the trailing-single-character length guard, so a lone out-of-range byte (`"z"`, `"....z"`) now reports `ErrGEDCOMBlobChar` instead of `ErrGEDCOMBlobLength`. Previously the length guard ran first and the byte was never validated, leaving a hole in the "every invalid-character failure is detectable via `errors.Is`" contract from item 2. The godoc states the precedence explicitly.

Also: the redundant `len(cleaned) == 1` branch is gone. It was unreachable as distinct behavior — for a single character `fullGroups` is 0 and `trailing` is 1, so the `trailing == 1` guard already returned the same sentinel — and it gave one failure class ("`.`" and "`.....`" are the same mistake) two different messages on the exported surface. The surviving "trailing single character cannot encode a full byte" wording is the more accurate of the two.

## Related issues

Closes #1140

## Review focus

API surface: the exported function name, the three sentinel error names, the godoc contract, and the exact bytes of the error messages — all free to change now and awkward to change after merge, with #1141 about to add a second caller.

Note that the decoder body is **not** a verbatim move: see "Behavior changes made while exporting" above. The widened whitespace set genuinely changes what input is accepted (strictly widening), and the CHANGELOG describes it as such.

## Testing

- Tests moved to `go-glx/gedcom_blob_test.go` and extended: `errors.Is` assertions for all three sentinels (empty, whitespace-only, single-char, 5-char, out-of-range in full and trailing groups), boundary cases at `'-'`/`'n'`, a known-value case (`"mmmm"` → `0xFF 0xFF 0xFF`), a whitespace case covering all six stripped bytes, trailing-group byte values, and a case pinning the `(0xFF)` hex rendering in the invalid-character message for both the full-group and trailing-group sites.
- `make test` passes; `go build ./...` and `go vet ./...` are clean.
- End-to-end: built the CLI and imported a GEDCOM 5.5.1 file with a `BLOB` record — the decoded bytes land in `media/files/blob-media-1.jpg` as before.

## Breaking changes

None for the CLI, and nothing to migrate for library consumers — `DecodeGEDCOMBlob` and its sentinels are new public API in this PR, so the renames are not visible outside it.

The `glx` binary's user-visible behavior changes in two small ways, both improvements: BLOBs padded with tabs, vertical tabs, or form feeds now import instead of erroring, and the invalid-character warning is reworded to include the sentinel's range hint and the offending byte in hex. The CLI package loses its unexported decoder and three `Err*` vars, which were not importable (package main).


go-glx/gedcom_blob.go
```diff
@@ -0,0 +1,97 @@
...
+	ErrEmptyBlobData     = errors.New("empty BLOB data")
+	ErrInvalidBlobLength = errors.New("invalid BLOB data length")
+	ErrInvalidBlobChar   = errors.New("invalid BLOB character (must be in range '.' to 'm')")
+)
+
+// DecodeGEDCOMBlob decodes GEDCOM 5.5.1 BLOB-encoded text to raw bytes.
+// GEDCOM BLOB uses a custom encoding where each character's value minus 0x2E ('.')
+// gives a 6-bit value. Groups of 4 characters encode 3 bytes; a trailing group of
+// 2 or 3 characters encodes 1 or 2 bytes. Whitespace and newlines are stripped
+// before decoding.
```
isaacschepp: The doc contract is broader than the implementation. Only `\n`, `\r`, and space are stripped — a tab survives cleaning and then fails character validation:

```
DecodeGEDCOMBlob("..\t..")
  -> invalid BLOB character (must be in range '.' to 'm') at position 2: '\t'
```

That was harmless while the decoder was CLI-internal, but as exported API the godoc is the contract a consumer codes against, and "whitespace ... stripped" reads as "all whitespace". Two ways out; either is fine, but the doc and the code should agree:

- Narrow the doc to what it does: `Spaces, carriage returns, and line feeds are stripped before decoding.`
- Or widen the code to match the doc, e.g. `strings.Map` dropping `unicode.IsSpace` runes (also covers `\t`, `\v`, `\f`, NBSP).

Worth a moment's thought rather than defaulting to the doc fix: GEDCOM 5.5.1 BLOB payloads arrive as `CONT` line values, and a lenient exporter emitting a stray tab is exactly the kind of thing an importer is expected to tolerate. If you widen the code, that also subsumes the pre-existing gap rather than documenting it.

go-glx/gedcom_blob.go
```diff
@@ -0,0 +1,97 @@
...
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// BLOB decoding errors
+var (
+	ErrEmptyBlobData     = errors.New("empty BLOB data")
+	ErrInvalidBlobLength = errors.New("invalid BLOB data length")
+	ErrInvalidBlobChar   = errors.New("invalid BLOB character (must be in range '.' to 'm')")
```
isaacschepp: Now that the full-group path wraps this sentinel too (the deliberate consistency fix), the range hint lands in the middle of every message:

```
invalid BLOB character (must be in range '.' to 'm') at position 2: 'z'
```

The position and the offending byte — the two things a user actually needs — are pushed behind a parenthetical. Since both call sites use the identical `"%w at position %d: %q"` format, moving the hint into the wrap reads better and keeps `errors.Is` working exactly as it does now:

```suggestion
	ErrInvalidBlobChar   = errors.New("invalid BLOB character")
```

with both wrap sites becoming:

```go
return nil, fmt.Errorf("%w at position %d: %q (must be in range '.' to 'm')", ErrInvalidBlobChar, i+j, char)
```

giving `invalid BLOB character at position 2: 'z' (must be in range '.' to 'm')` — which is also what the old CLI message looked like, so the CLI's user-facing warning stops being a regression rather than becoming one. Raising it now because the sentinel's `Error()` string is part of the exported surface this PR is freezing.

go-glx/gedcom_blob.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+// DecodeGEDCOMBlob decodes GEDCOM 5.5.1 BLOB-encoded text to raw bytes.
+// GEDCOM BLOB uses a custom encoding where each character's value minus 0x2E ('.')
+// gives a 6-bit value. Groups of 4 characters encode 3 bytes; a trailing group of
+// 2 or 3 characters encodes 1 or 2 bytes. Whitespace and newlines are stripped
+// before decoding.
+func DecodeGEDCOMBlob(blobText string) ([]byte, error) {
+	// Strip whitespace and newlines
+	cleaned := strings.NewReplacer("\n", "", "\r", "", " ", "").Replace(blobText)
```
isaacschepp: `strings.NewReplacer` builds and compiles its trie on every call. `go-glx/CLAUDE.md` calls out this exact pattern under Performance Profiling ("Map literals in function bodies allocate every call — use package-level vars"), and it matters more here than it did in the CLI: #1141 will call this once per BLOB media record across a whole GEDZIP, not once per invocation.

Hoist it:

```go
// blobWhitespace strips the line breaks and spaces GEDCOM CONT lines introduce.
var blobWhitespace = strings.NewReplacer("\n", "", "\r", "", " ", "")
```

and use `blobWhitespace.Replace(blobText)`. `strings.Replacer` is safe for concurrent use, so a package-level var is fine for a library. (Moot if you take the `strings.Map` route in the doc comment thread.)

go-glx/gedcom_blob_test.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	// Trailing 2 chars (valid — produces 1 byte)
+	decoded4, err := DecodeGEDCOMBlob("..")
+	if err != nil {
+		t.Fatalf("Decode of 2 chars failed: %v", err)
+	}
+	if len(decoded4) != 1 {
+		t.Errorf("Expected 1 byte from 2 chars, got %d", len(decoded4))
+	}
```
isaacschepp, line 77: These two only assert the output *length*, so they'd pass unchanged if the trailing-group bit math were wrong — the all-`'.'` inputs decode to zero bytes, which is also what a broken shift would plausibly produce. The full-group case got a real known-value check with `"mmmm"`; the trailing paths are the fiddlier code and have none.

I verified the current values by hand, so these are correct today and can be asserted directly:

```go
// "mmm" → 2 bytes, "mm" → 1 byte, all bits set
if b, err := DecodeGEDCOMBlob("mmm"); err != nil || !bytes.Equal(b, []byte{0xFF, 0xFF}) {
    t.Errorf("Expected [0xFF 0xFF] from 3 chars, got %v (err %v)", b, err)
}
if b, err := DecodeGEDCOMBlob("mm"); err != nil || !bytes.Equal(b, []byte{0xFF}) {
    t.Errorf("Expected [0xFF] from 2 chars, got %v (err %v)", b, err)
}
```

A mixed-value case would be stronger still (all-zero and all-ones both survive a swapped shift direction), e.g. one input where the two trailing sextets differ, so the `(b1<<2)|(b2>>4)` split is actually pinned.

go-glx/gedcom_blob_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// Out-of-range character in a full group ('z' > 'm')
+	if _, err := DecodeGEDCOMBlob("..z."); !errors.Is(err, ErrInvalidBlobChar) {
+		t.Errorf("Expected ErrInvalidBlobChar for out-of-range char in full group, got %v", err)
+	}
+
+	// Out-of-range character in a trailing group
+	if _, err := DecodeGEDCOMBlob("....z."); !errors.Is(err, ErrInvalidBlobChar) {
+		t.Errorf("Expected ErrInvalidBlobChar for out-of-range char in trailing group, got %v", err)
+	}
```
isaacschepp, line 165: Both cases probe the upper bound (`'z'` > `'m'`). The lower bound — `char < '.'` — is never exercised, in either the full-group or trailing branch. That half of the condition could be deleted and the suite would stay green.

`'-'` is 0x2D, exactly one below `'.'`, so it pins the boundary rather than just testing "some low byte":

```go
// Character just below the valid range ('-' is 0x2D, one below '.')
if _, err := DecodeGEDCOMBlob("-..."); !errors.Is(err, ErrInvalidBlobChar) {
    t.Errorf("Expected ErrInvalidBlobChar for below-range char in full group, got %v", err)
}
if _, err := DecodeGEDCOMBlob("....-."); !errors.Is(err, ErrInvalidBlobChar) {
    t.Errorf("Expected ErrInvalidBlobChar for below-range char in trailing group, got %v", err)
}
```

Worth adding the upper-boundary accept case too — `"mmmm"` in the happy-path test already covers `'m'` being valid, so the pair `'m'` accepted / `'n'` rejected is fully pinned once `'n'` shows up somewhere.

isaacschepp (changes requested):
## Review — export GEDCOM 5.5.1 BLOB decoder

The move is well-scoped and the pure-function/no-I/O rule in `go-glx/CLAUDE.md` is respected. Verified locally on the branch (Go 1.26.1):

- `go vet ./go-glx/... ./glx/...` clean
- `go test ./glx/... ./go-glx/...` — both packages pass
- `golangci-lint run --new-from-rev origin/main` — no findings in `gedcom_blob.go`, `gedcom_blob_test.go`, `media_copy.go`, or `glx/errors.go`
- No dangling references to `decodeGEDCOMBlob` or the three `Err*` vars remain in `package main`; `glx/media_copy.go` still uses `fmt`/`strings` elsewhere so no unused-import fallout

**Correctness of the moved body** — checked independently, not just diffed:

- Bit math is right. `"mmmm"` (four 63s) → `0xFF 0xFF 0xFF`; `"mm"` → `0xFF`; `"mmm"` → `0xFF 0xFF`. Trailing 3-char group correctly discards the unused low bits of the third sextet.
- Capacity `len(cleaned)*3/4` is *exact* for every residue class (4k→3k, 4k+2→3k+1, 4k+3→3k+2), so no reallocation and no over-allocation.
- No out-of-bounds risk: full-group indices are bounded by `fullGroups <= len(cleaned)`, trailing indices by `fullGroups+trailing-1 == len(cleaned)-1`. Safe on untrusted input; no panic path, no unbounded allocation beyond 3/4 of input.
- Validation runs before any byte is emitted in each group, so a failure never returns partial output.
- Local sentinel declaration next to the feature (rather than `go-glx/errors.go`) matches the `census.go` precedent — fine.
- `DecodeGEDCOMBlob` over `DecodeBlob` is the right call, and `Blob` (not `BLOB`) matches `MediaSourceBlob`/`BlobData`.

Nothing blocking correctness. The comments below are all about the fact that this PR *freezes* an API surface — the error text and the doc contract are the parts that get expensive to change later, so they're worth a second pass now.

### Findings

1. `gedcom_blob.go:33` — the new doc sentence overstates what is stripped (tabs are not).
2. `gedcom_blob.go:27` — the sentinel's parenthetical now lands mid-sentence in every wrapped message.
3. `gedcom_blob.go:37` — `strings.NewReplacer` rebuilt on every call.
4. `gedcom_blob_test.go:66` — trailing-group tests assert length only, not bytes.
5. `gedcom_blob_test.go:111` — only the upper range bound is covered; nothing tests below `'.'`.

### Not blocking, optional

`go-glx/doc.go`'s "This package provides" bullet list doesn't mention BLOB decoding. Since the whole point of #1140 is discoverability for library consumers, one bullet there would help more than the godoc on the function itself.

### CHANGELOG

Entry is accurate and correctly filed under Unreleased → Added, including the error-message change. No `### Changed` entry needed since the CLI is the only consumer and its behavior is otherwise identical.

go-glx/gedcom_blob.go
isaacschepp: Addressed in 45811155. Took the widen-the-code route rather than narrowing the doc: `blobWhitespace` now strips `\t`, `\v` and `\f` alongside space, `\r` and `\n`, so a lenient exporter's stray tab decodes instead of erroring. Stayed with `strings.Replacer` rather than `strings.Map`/`unicode.IsSpace` for the reason noted in the thread's alternative: `strings.Map` is rune-oriented and would fold invalid UTF-8 bytes into U+FFFD, which is 3 bytes wide and would shift every subsequent position reported by `ErrInvalidBlobChar`. Byte-oriented stripping keeps the offsets honest on malformed input. `TestDecodeGEDCOMBlobWhitespace` covers each whitespace byte as a group separator, and the whitespace-only case in `TestDecodeGEDCOMBlobErrors` now uses all six.

go-glx/gedcom_blob.go
isaacschepp: Addressed in 45811155. Sentinel is now `errors.New("invalid BLOB character")` and both wrap sites use `"%w at position %d: %q (must be in range '.' to 'm')"`, giving `invalid BLOB character at position 2: 'z'` — position and offending byte first, hint last, matching the pre-export CLI wording. `errors.Is` still matches on both the full-group and trailing-group paths.

go-glx/gedcom_blob.go
isaacschepp: Addressed in 45811155. Hoisted to a package-level `blobWhitespace` var with a comment recording why it is byte-oriented. `strings.Replacer` is safe for concurrent use, so this is fine for a library.

go-glx/gedcom_blob_test.go
isaacschepp: Addressed in 45811155. Added `TestDecodeGEDCOMBlobTrailingValues`, a table with mixed sextets so a swapped shift direction can't pass: `"m."` -> `0xFC`, `".m"` -> `0x03`, `"m.m"` -> `0xFC 0x0F`, `".m."` -> `0x03 0xF0`, plus the all-ones `"mm"`/`"mmm"` cases. Values were computed by hand first and matched the implementation on the first run.

go-glx/gedcom_blob_test.go
isaacschepp: Addressed in 45811155. Both boundaries are now pinned on both code paths: `'-'` (0x2D, one below `'.'`) and `'n'` (0x6E, one above `'m'`), each in a full group and a trailing group. `'m'` itself staying valid is covered by the existing `"mmmm"` case, which the new comment points at.

go-glx/gedcom_blob.go
```diff
@@ -0,0 +1,109 @@
...
+	result := make([]byte, 0, len(cleaned)*3/4)
+	fullGroups := (len(cleaned) / 4) * 4 //nolint:mnd // 4 chars per group
+
+	for i := 0; i < fullGroups; i += 4 {
+		// Validate each character is in valid GEDCOM BLOB range (0x2E '.' to 0x6D 'm')
+		// This gives 6-bit values (0-63) after subtracting 0x2E
+		for j := range 4 {
+			char := cleaned[i+j]
+			if char < '.' || char > 'm' {
+				return nil, fmt.Errorf("%w at position %d: %q (must be in range '.' to 'm')", ErrInvalidBlobChar, i+j, char)
```
isaacschepp, line 68: `%q` on a `byte` formats the value as a *rune*, so any offending byte >= 0x80 is reported as the Latin-1 character with that code point rather than as the byte actually present in the input. I checked this against the toolchain rather than from memory:

```
byte 0xFF -> %q = 'y-diaeresis'   (U+00FF)
byte 0xC3 -> %q = 'A-tilde'       (U+00C3)
byte 0x80 -> %q = '\u0080'
byte 0x0B -> %q = '\v'
```

This matters specifically because of the rationale the new comment on `blobWhitespace` states four lines up: byte-orientation was chosen *"so that invalid UTF-8 in a malformed BLOB passes through untouched and is reported at its true byte offset instead of as U+FFFD."* The offset half of that promise holds. The rendering half does not — a BLOB carrying a UTF-8 lead byte or arbitrary binary junk produces a message naming a character that does not appear in the input, which is the same class of confusion the U+FFFD substitution would have caused.

The realistic case is a mis-encoded or truncated media BLOB, which is exactly when a consumer is squinting at the error to work out what the exporter emitted. Showing the hex removes the ambiguity:

```go
return nil, fmt.Errorf("%w at position %d: %q (0x%02X) (must be in range '.' to 'm')", ErrInvalidBlobChar, i+j, char, char)
```

Applies identically to the trailing-group site on line 86. If you'd rather not carry both renderings, hex alone is the one that can't lie.

go-glx/gedcom_blob.go
```diff
@@ -0,0 +1,109 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// BLOB decoding errors
+var (
+	ErrEmptyBlobData     = errors.New("empty BLOB data")
```
isaacschepp, line 25: These three become public API on merge, so the naming rationale in the PR description is worth applying to them too. That rationale is:

> Named `DecodeGEDCOMBlob` rather than `DecodeBlob`: the package is `glx`, so `glx.DecodeBlob` would read as "decode a GLX blob".

That argument is entirely right, and it applies verbatim to `glx.ErrInvalidBlobChar` — at a call site that reads as an invalid character in a GLX blob, which is not a thing that exists. The function got disambiguated; the errors it returns did not.

There's also in-package precedent for the domain prefix. `go-glx/census.go` declares its four exported sentinels co-located with the feature (which is why I think the placement here is fine, not a deviation) and prefixes every one of them:

```go
ErrCensusTemplateRequired
ErrCensusYearRequired
ErrCensusLocationRequired
ErrCensusHouseholdMembersMissing
```

`ErrGEDCOMBlobEmpty` / `ErrGEDCOMBlobLength` / `ErrGEDCOMBlobChar` would match both that pattern and `DecodeGEDCOMBlob`. Renaming is free right now and a deprecation cycle later — and #1141 is about to add the second caller, which is the point where the names start to stick.

go-glx/gedcom_blob.go
```diff
@@ -0,0 +1,109 @@
...
+// whole number of bytes, and [ErrInvalidBlobChar] when a character falls outside
+// the '.' to 'm' range.
+func DecodeGEDCOMBlob(blobText string) ([]byte, error) {
+	cleaned := blobWhitespace.Replace(blobText)
+
+	if len(cleaned) == 0 {
+		return nil, ErrEmptyBlobData
+	}
+
+	if len(cleaned) == 1 {
```
isaacschepp, line 55: This branch is unreachable as *distinct* behavior — it only changes the message text, never the outcome.

For `len(cleaned) == 1`: `fullGroups` is `(1/4)*4 == 0`, the full-group loop doesn't execute, and `trailing` is `1 - 0 == 1`, so the `trailing == 1` guard on line 92 already catches it and returns the same `ErrInvalidBlobLength`. Deleting lines 55-57 changes exactly one thing: the message becomes `invalid BLOB data length: 1 (trailing single character cannot encode a full byte)` instead of `invalid BLOB data length: 1 (minimum 2 characters required)`.

So the exported API currently has two different phrasings for one failure class, distinguished only by whether the stray character happened to land alone or after some full groups — `"."` and `"....."` are the same mistake and should read the same way. The trailing-path wording is the more accurate of the two (it explains *why* one character can't work), and it already handles both.

Pre-existing, carried over from the CLI, and harmless while it was internal. Worth collapsing now that the message is something consumers may match on or surface to users.

go-glx/gedcom_blob.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+// BLOB decoding errors
+var (
+	ErrEmptyBlobData     = errors.New("empty BLOB data")
+	ErrInvalidBlobLength = errors.New("invalid BLOB data length")
+	ErrInvalidBlobChar   = errors.New("invalid BLOB character")
+)
+
+// blobWhitespace strips the ASCII whitespace that GEDCOM CONT/CONC line
```
isaacschepp, line 30: The attribution here isn't accurate, and it obscures the actual reason the set was widened.

CONT/CONC line wrapping introduces line terminators — `\r` and `\n`. It does not emit tabs, vertical tabs, or form feeds; the space separating the tag from its value is consumed by the line parser before `BlobData` is assembled (`go-glx/gedcom_media.go:275`). So `\t`, `\v` and `\f` are in this replacer for a different reason than the comment gives.

The real reason is the one from the earlier thread — tolerating a lenient or pretty-printing exporter that padded the BLOB payload. That's a defensible call and strictly widening (all three bytes are below `'.'`, so they used to hard-fail), but a future reader who trusts this comment will conclude the parser can produce a tab, go looking for where, and find nothing. Something like:

```go
// blobWhitespace strips the ASCII whitespace GEDCOM CONT/CONC line wrapping
// introduces (\r, \n), plus the padding bytes lenient exporters have been
// seen to insert (space, \t, \v, \f). All are outside the '.'-'m' range, so
// stripping them widens what decodes without changing any existing decode.
```

The second sentence of the existing comment — the byte-vs-rune rationale — is good and worth keeping as-is.

go-glx/gedcom_blob.go
```diff
@@ -0,0 +1,109 @@
...
+	" ", "", "\t", "", "\n", "", "\v", "", "\f", "", "\r", "",
+)
+
+// DecodeGEDCOMBlob decodes GEDCOM 5.5.1 BLOB-encoded text to raw bytes.
+// GEDCOM BLOB uses a custom encoding where each character's value minus 0x2E ('.')
+// gives a 6-bit value. Groups of 4 characters encode 3 bytes; a trailing group of
+// 2 or 3 characters encodes 1 or 2 bytes. ASCII whitespace (spaces, tabs and line
+// breaks) is stripped before decoding.
+//
+// The returned error wraps [ErrEmptyBlobData] when the text holds no encoded
```
isaacschepp, line 44: Small contract inaccuracy in godoc, worth fixing since this paragraph *is* the contract now.

"The returned error wraps `[ErrEmptyBlobData]`" — the empty case returns the sentinel bare (`return nil, ErrEmptyBlobData`, line 52), not wrapped. The other two genuinely are wrapped via `%w`. `errors.Is` works either way, so no consumer breaks, but a consumer reading this could reasonably expect `errors.Unwrap` to yield something, or expect the message to carry context the way the length and character errors do.

"The returned error matches [ErrEmptyBlobData] ... under errors.Is" covers all three accurately without committing to a wrapping depth you may not want to promise.

Same paragraph, line 41-42: *"ASCII whitespace (spaces, tabs and line breaks) is stripped"* is now the exported contract for whitespace handling. It reads as illustrative but is actually exhaustive — `\v` and `\f` are stripped too and aren't naturally covered by "tabs and line breaks". Either name all six or say "ASCII whitespace" without the parenthetical, so the doc and `blobWhitespace` can't drift apart.

isaacschepp (changes requested):
Re-review after 45811155. All five earlier threads are genuinely addressed — the trailing-group value table, the boundary characters on both paths, the hoisted `Replacer`, and the reworded `ErrInvalidBlobChar` all do what the replies claim.

**Verification I ran independently of CI:** decoded the bit math by hand for every case in `TestDecodeGEDCOMBlobTrailingValues` (`(b1<<2)|(b2>>4)` and `(b2<<4)|(b3>>2)` under `byte` truncation) — all six expectations are right. Confirmed the capacity hint `len(cleaned)*3/4` is exact, not just sufficient, for every trailing size: `(4k+t)*3/4 == 3k + floor(3t/4)`, which is `3k`, `3k+1`, `3k+2` for `t` = 0, 2, 3. Ran `go test ./glx/... ./go-glx/...` on the branch — both packages pass. `strings` is still used elsewhere in `glx/media_copy.go`, so removing the decoder leaves no orphaned import.

**The move itself is right.** Pure `string -> ([]byte, error)`, no I/O, fully compliant with [ADR-0006](https://github.com/genealogix/glx/blob/main/docs/decisions/0006-go-glx-library-pure.md). Declaring the sentinels in `gedcom_blob.go` rather than `go-glx/errors.go` also matches existing precedent — `go-glx/census.go` co-locates its four exported sentinels the same way, so this is the convention, not a deviation from it.

Nothing below is a correctness defect. The reason I'm not approving outright is that four of the five are about the **exported surface** — names, godoc contract, and the bytes an error message shows a consumer — and this PR is the last moment they are free to change. `DecodeGEDCOMBlob`, three sentinels, and their godoc become library API the moment this merges.

### One thing to fix in the PR description before merge

The **Review focus** and **Breaking changes** sections are stale after 45811155. They still say the decoder body is *"moved verbatim apart from the one error-wrapping change described above"* and that CLI behavior is *"unchanged except the slightly reworded invalid-character error message."*

Neither holds any more. Widening `blobWhitespace` to strip `\t`, `\v` and `\f` is a real change in what input the decoder accepts: those three bytes are all below `'.'` (0x09, 0x0B, 0x0C vs 0x2E), so a BLOB containing them previously failed with `ErrInvalidBlobChar` and now decodes. It is strictly widening — no input that decoded before decodes differently now, which is why I'm not treating it as a behavior risk — but it reached the code as a response to review, and the description was never updated to match. The CHANGELOG entry describes it correctly; the PR body contradicts it, and it is the PR body that tells a reviewer where *not* to look.

### Summary

| # | File | Concern |
|---|------|---------|
| 1 | `go-glx/gedcom_blob.go:68` | `%q` on a `byte` misreports the offending byte for values >= 0x80 |
| 2 | `go-glx/gedcom_blob.go:25` | Sentinel names drop the `GEDCOM` qualifier the function name deliberately carries |
| 3 | `go-glx/gedcom_blob.go:55` | `len(cleaned) == 1` branch is redundant with the `trailing == 1` path |
| 4 | `go-glx/gedcom_blob.go:30` | Comment attributes the stripped bytes to CONT/CONC wrapping, which doesn't introduce them |
| 5 | `go-glx/gedcom_blob.go:44` | Godoc says the empty-blob error "wraps" a sentinel it returns bare |

Items 3-5 are cheap enough to fold into one commit; 1 and 2 are the ones I'd actually like resolved before this becomes public API.

go-glx/gedcom_blob.go
noahbjohnson: Addressed in 0b812d1a. Both invalid-character sites now render the offending byte as `%q (0x%02X)`, so a byte >= 0x80 is reported as the byte actually in the input alongside the (possibly misleading) rune rendering. `TestDecodeGEDCOMBlobCharErrorReportsByte` pins the `(0xFF)` hex in the message on both the full-group and trailing-group paths.

go-glx/gedcom_blob.go
noahbjohnson: Addressed in 0b812d1a. Renamed to `ErrGEDCOMBlobEmpty` / `ErrGEDCOMBlobLength` / `ErrGEDCOMBlobChar`, matching both `DecodeGEDCOMBlob` and the `census.go` sentinel-prefix precedent. The godoc, tests, CLI caller, CHANGELOG, and PR description all use the new names.

go-glx/gedcom_blob.go
noahbjohnson: Addressed in 0b812d1a. The `len(cleaned) == 1` branch is gone; the `trailing == 1` guard is the single path for that failure class, so `"."` and `"....."` now produce the same message (the more accurate trailing-group wording). Follow-up in 12964e1e: trailing characters are now validated *before* that guard, so a lone out-of-range byte reports `ErrGEDCOMBlobChar` rather than the length error — closing the `errors.Is` contract hole Copilot flagged.

Issue #931, opened 2026-05-27
releases: enable GitHub Immutable Releases (GA Oct 2025) to lock tag and asset refs after publish


isaacschepp:
## Summary

GitHub [Immutable Releases became GA on October 28, 2025](https://github.blog/changelog/2025-10-28-immutable-releases-are-now-generally-available/). When enabled, release assets and the underlying Git tag are frozen after publish — neither can be overwritten, deleted, or have their target SHA changed without first un-immutabilizing the release. glx does not have this enabled (verifiable via repo Settings → Code, planning, and automation → General → Releases, or `gh api repos/genealogix/glx --jq '.is_template'`-adjacent endpoints).

## Problem

The `release.yml` workflow signs `checksums.txt` with cosign keyless (per [SECURITY-POSTURE.md](https://github.com/genealogix/glx/blob/main/SECURITY-POSTURE.md) L3 row), and signing the checksum manifest transitively binds every artifact via SHA-256. That protects against **content modification** of release artifacts.

It does **not** protect against **tag rewriting** — an attacker with write access (or temporary write access via a compromised PAT) can:

1. Push a new commit `Z'` to the repository.
2. Force-update tag `vX.Y.Z` to point at `Z'`.
3. Delete the existing release and re-create one at the new tag.
4. The new release ships unsigned (or with a fresh attacker-controlled signature).
5. Downstream consumers who pinned to the tag (rather than the original SHA + signature bundle) pick up the malicious release on next pull.

This is exactly the **Laravel-Lang attack pattern (May 22, 2026)**: attacker with push access rewrote every Git tag across multiple Composer packages in 15 minutes ([writeup](https://lyrie.ai/research/research/github-actions-systemic-supply-chain-pwn-requests)). The tj-actions/changed-files attack (CVE-2025-30066, Mar 2025) used a similar primitive — force-rewriting every version tag v1 through v45.0.7 to a malicious commit.

Immutable Releases mitigate this end-to-end at the GitHub side: once a release is published, its tag is locked to the original SHA, assets are read-only, and an [auto-issued release attestation](https://docs.github.com/en/actions/how-tos/secure-your-work/use-artifact-attestations/about-artifact-attestations) records the immutability. Even an attacker with full repo write access cannot tamper with an existing release.

## Proposed Solution

**Enable Immutable Releases on the repository** via Settings → Code, planning, and automation → General → Releases → "Immutable Releases."

[...]

Issue #858, opened 2026-05-15
Discussion: replace post-hoc compact-changelog with file-per-change tooling (evaluate changie for 2026)


isaacschepp:
## Summary

`compact-changelog.md` exists because agentic edits create duplicate `### Added` sections, self-cancelling Added/Removed pairs, and follow-up Fixed entries that should fold into the original Added. **File-per-change tools eliminate this class of problem at the source by writing each change to its own file**, then assembling at release time. The compact-changelog command is, at best, post-hoc cleanup of a problem that need not exist.

This is a Discussion, not a Bug or Enhancement — it's a strategic question: keep the cleanup tool, or remove the underlying source of mess.

## The 2026 landscape

| Tool | Language | Format | KaC interop | GoReleaser support | Fit for GLX |
|---|---|---|---|---|---|
| **changie** (`miniscruff/changie`) | Go binary | YAML fragments in `.changes/unreleased/` | Templates to any format including KaC | Official integration (`changie.dev/integrations/goreleaser`) | **Strong** |
| towncrier | Python | reST/MD fragments | Markdown emit since 22.x | Manual | Medium (Python dep) |
| changesets | Node | MD fragments | npm-monorepo-centric | Manual | Weak (wrong ecosystem) |
| git-cliff | Rust binary | n/a — generates from commits | Yes | Manual | Strong but a different philosophy |
| release-please | Action | Release-PR pattern | Multi-lang incl. Go | Replaces GoReleaser changelog | Heavy, opinionated |
| GoReleaser-Pro AI changelog | Go (paid) | Auto from commits + LLM polish | Auto | Native | Paid license needed |

Changie's happy path:

```bash
# Each agent/contributor, per change, runs:
changie new
# Interactive prompt: kind (Added/Changed/...), body, optional refs

# At release time:
changie batch v0.0.0-beta.11    # collect fragments → CHANGELOG.md section
changie merge                    # write into CHANGELOG.md preserving the rest
```

Fragments live in `.changes/unreleased/<random>.yaml`. Two agents adding entries the same minute create two separate files; no merge conflicts, no duplicate `### Added` headers, no need to compact.

## What this would change for GLX

**Removes:**

- The merge-conflict surface on `CHANGELOG.md` that `auto-resolve-conflicts.yml` exists to handle.
- The "Feature branch hygiene: `git checkout main -- CHANGELOG.md`, then re-add branch entries" instruction in `CLAUDE.md`.
[...]

PR #1180, opened 2026-08-23
ci: Harden release.yml and mark prereleases correctly

### LINKED ISSUE #1007: release.yml: -beta tags publish as full (potentially "Latest") GitHub Releases — add release.prerelease: auto to .goreleaser.yml
## Summary

`release.yml` deliberately triggers on beta tags (line 7: `- 'v[0-9]+.[0-9]+.[0-9]+-beta*'`), but `.goreleaser.yml` has **no `release:` stanza at all**, so GoReleaser's `release.prerelease` field uses its default of `false`. Auto-detection of a semver prerelease suffix only happens when `release.prerelease: auto` is explicitly set. Consequently a tag like `v0.0.0-beta.12` is published as a normal, **non-prerelease** GitHub Release rather than being flagged `Pre-release`.

## This is happening right now (live evidence)

```
$ gh api repos/genealogix/glx/releases/latest --jq '.tag_name +" prerelease="+ (.prerelease|tostring)'
v0.0.0-beta.11 prerelease=false
```

`GET /releases/latest` returns "the most recent non-prerelease, non-draft release", so **`v0.0.0-beta.11` is currently the repo's official "Latest release"** and is what install scripts / `gh release download` (no `--pattern`/tag) / the API resolve to. Worse, the flag *regressed*: `beta.0` is `prerelease=true`, but every release from `beta.1` through `beta.11` is `prerelease=false`:

```
v0.0.0-beta.11 .. v0.0.0-beta.1   prerelease=false
v0.0.0-beta.0                     prerelease=true
```

So betas were once correctly flagged and silently stopped being flagged. The dedicated `-beta` trigger glob shows clear intent to treat betas as a separate channel; the GoReleaser side does not honor it. The Discord announcement (release.yml:64) also advertises every beta as "A new version of the GLX CLI has been released" with no beta qualifier.

The blast radius is bounded today only because there is no stable `vX.Y.Z` line yet — but the moment a stable release is cut alongside betas, a freshly-pushed beta can out-sort and claim the Latest badge.

## Evidence

- `.github/workflows/release.yml:7` — `- 'v[0-9]+.[0-9]+.[0-9]+-beta*'` (beta trigger present; the workflow runs `args: release`, a real tagged publish, not a snapshot).
- `.goreleaser.yml` — no `release:` key anywhere (file ends with the `changelog:` block; grep for `release:|prerelease|draft` returns nothing).
- GoReleaser default is `prerelease: false`; `make_latest` defaults to `true`.

## Recommended fix

Add a top-level `release:` block to `.goreleaser.yml`:

```yaml
release:
  prerelease: auto   # marks -alpha/-beta/-rc tags as GitHub pre-releases
```

With `auto`, GoReleaser inspects the tag's semver prerelease segment and flags `-beta` (and future `-rc`/`-alpha`) builds as pre-releases, excluding them from `/releases/latest` and the Latest badge. Optionally also gate the Discord copy on the prerelease state so betas are not announced as plain releases.

Secondary observation worth a one-line doc comment: the trigger only whitelists `-beta*`; `-rc`/`-alpha` tags match neither glob (`v[0-9]+.[0-9]+.[0-9]+` is anchored with no trailing wildcard), so they would not trigger a release at all. `prerelease: auto` is forward-looking insurance once those are added.

## Cosign contract safety

Safe. `release:` and `signs:` are independent top-level keys in `.goreleaser.yml`. The fix changes only the GitHub Release object's metadata (prerelease flag / Latest eligibility); it does not alter the signed artifact set (`checksums.txt`), the `cosign sign-blob --bundle` args, the `.sigstore.json` output format, or the OIDC keyless identity bound to `release.yml@refs/tags/`. `cosign verify-blob --certificate-identity-regexp '^https://github\.com/genealogix/glx/\.github/workflows/release\.yml@refs/tags/'` continues to match unchanged. Only operational note: downstream automation currently relying on `/releases/latest` returning a beta will, correctly, begin skipping prereleases.

## References

- GoReleaser Releases customization: https://goreleaser.com/customization/release/
- Origin of alpha/beta/rc auto-detection: https://github.com/goreleaser/goreleaser/issues/667
- GitHub `/releases/latest` semantics: https://docs.github.com/en/rest/releases/releases#get-the-latest-release

---
*Found during a deep 2026 best-practice review of `release.yml`. Verified live against the GitHub Releases API.*

### LINKED ISSUE #1008: release.yml: add a concurrency group so concurrent/moved-tag pushes cannot race the signing+publish job
## Summary

`release.yml` has no `concurrency:` block, while 11 of the repo's other workflows declare one. The release workflow is the one that does irreversible, externally-visible work — uploads GitHub Release assets, keyless-signs `checksums.txt` with cosign, and posts to Discord — yet it has no serialization.

The repo's own `.github/CLAUDE.md` documents a tag-move hotfix flow (`git tag -d X && git push --delete origin X && git tag X && git push origin X`). A delete-then-repush, or two tag pushes in quick succession, can launch two release runs that race on the same GitHub Release: GoReleaser returns a 422 `already_exists` on the loser's asset upload, uploads can interleave, and both runs fire a Discord announcement.

Two open issues (#950 issue-templates-drift.yml, #995 lychee.yml) already treat "lone workflow without a concurrency group" as a fileable defect, establishing the repo direction; the higher-stakes release workflow warrants its own.

## Evidence

- `.github/workflows/release.yml` — no top-level or job-level `concurrency:` key.
- 11 peers declare one (e.g. `scorecard.yml:13`, `security.yml:14`, `lint.yml:25`, `dependency-review.yml:9`); `auto-update-branches.yml`/`auto-resolve-conflicts.yml` use a shared `branch-maintenance` group with `cancel-in-progress: false`.
- `.github/CLAUDE.md` "Release pipeline" section documents the delete-and-repush tag-move procedure.

## Recommended fix

Add at workflow top level (after `permissions:` on line 9):

```yaml
concurrency:
  group: release-${{ github.ref }}
  cancel-in-progress: false
```

Use `cancel-in-progress: false` (NOT `true`) — cancelling a half-uploaded/half-signed release mid-flight is worse than queueing, and `.github/CLAUDE.md` already warns that moving a tag after assets are uploaded orphans binaries and confuses `cosign verify-blob`. The group serializes per-tag (`github.ref` for a tag push is `refs/tags/vX.Y.Z`), so a moved-tag delete+repush groups correctly.

Honest scoping note: concurrency serializes but does not by itself *fix* the re-push-mid-run case — the queued loser can still hit `already_exists`. The value is eliminating interleaved uploads and the concurrent double-Discord fire. (Combine with the `prerelease`/validation issues for full release-resilience.)

## Cosign contract safety

Safe. `concurrency:` is a scheduling-only mechanism; it does not change the workflow file path, the triggering ref, the OIDC subject claim, or the `id-token: write` permission, so the `--certificate-identity-regexp` bound to `release.yml@refs/tags/` is untouched. Do **NOT** use `cancel-in-progress: true` — that could terminate a run between asset upload and cosign signing, leaving an unsigned/partially-signed release.

## References

- Control workflow concurrency: https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency
- GoReleaser release-upload `already_exists`: https://goreleaser.com/errors/release-upload

---
*Found during a deep 2026 best-practice review of `release.yml`.*

### LINKED ISSUE #1009: release.yml: pin the floating GoReleaser tool version (~> v2) to match the repo's pin-everything posture
## Summary

`release.yml:49` sets `version: '~> v2'`, a max-satisfying semver range that resolves to the **newest GoReleaser v2.x at run time**. Pinning the goreleaser-*action* SHA (line 47) does not pin the goreleaser *tool binary* it downloads. So re-running an old tag months later (e.g. via the moved-tag hotfix flow in `.github/CLAUDE.md`) can build with a newer GoReleaser whose archive naming / changelog filtering / output defaults differ — the produced artifacts are not guaranteed identical across time.

This is the **one floating version** among an otherwise exactly-pinned set on adjacent steps: `cosign-installer@v4.1.2` (line 42), `cosign-release: 'v3.0.6'` (line 44, commented "part of our release verification contract"), and the SHA-pinned goreleaser-action (line 47). It is a consistency/reproducibility hardening, not a bug — `~> v2` works correctly and the GoReleaser v2 major is backward-compatible within itself.

## Evidence

- `release.yml:49` — `version: '~> v2'`.
- Contrast: `release.yml:42` `@v4.1.2`, `:44` `cosign-release: 'v3.0.6'`, `:47` SHA-pinned action.

## Recommended fix

```yaml
        with:
          version: '~> v2.x.y'   # pin the tool to match repo posture; bump deliberately
          args: release
```

Prefer `~> v2.x.y` (patch-floating within a chosen minor) over an exact `v2.x.y` so security/bug patches within the pinned minor are still picked up. Add a trailing comment so it is not "tidied" back to `~> v2`.

**Operational caveat:** Dependabot's github-actions ecosystem updates action `uses:` SHAs but does **NOT** manage the `with: version:` input string — this pin must be bumped manually or via dedicated config (worth confirming against this repo's `dependabot.yml`).

Note: the *larger* cross-time reproducibility lever is actually the Go toolchain (`go-version: '1.26'` floats to the newest patch), tracked separately in #959. This issue is the GoReleaser-tool analogue.

## Cosign contract safety

Safe. The signing logic lives in `.goreleaser.yml` `signs:` and the `--bundle`/`.sigstore.json` format is pinned via `cosign-release: v3.0.6` (independent of the goreleaser tool). The OIDC identity is bound to the workflow path/ref, not the tool version. Pin to a version that still emits `${artifact}.sigstore.json`. Keep the goreleaser-action SHA pin (line 47) as-is.

## References

- goreleaser-action version input: https://github.com/goreleaser/goreleaser-action
- GoReleaser immutable releases: https://goreleaser.com/blog/goreleaser-v2.16/

---
*Found during a deep 2026 best-practice review of `release.yml`.*

### LINKED ISSUE #1010: release.yml: harden the Discord announce step — build JSON with jq, stage github.repository as env
## Summary

The Discord announce step (`release.yml:58-73`) has two robustness/consistency issues. (Note: the step's secret guard `if: success() && env.DISCORD_WEBHOOK != ''` with step-local `env` is **correct** and needs no change — step-level env is visible to that step's own `if:`, and secrets cannot be referenced in `if:` directly, so staging into env is the official pattern.)

1. **Brittle hand-built JSON.** `TAG="${GITHUB_REF_NAME}"` is spliced raw into a hand-written `\"title\": \"GLX ${TAG}\"` string. The `-beta*` trigger glob's trailing `*` matches any character except `/`, and git permits `"` in a ref name, so a tag like `v1.2.3-beta"x` is valid, triggers the step, and produces malformed JSON — `curl -sSf` then fails the step (a crafted `"` could even inject extra JSON fields into the payload). git **forbids** backslash and control chars in ref names, so the double-quote is the only reachable break character; tags are maintainer-pushed, so practical risk is low — this is brittleness, not an exploitable vector.

2. **Inline `${{ github.repository }}` in `run:`** (line 60). `github.repository` is a constrained `owner/repo` value (no injection vector), but inlining an expression into `run:` is inconsistent with the repo's own standing rule in `.github/CLAUDE.md` ("stage as `env:`, reference `\"$VAR\"`"); the runner already provides `$GITHUB_REPOSITORY`.

## Evidence

- `release.yml:59-60` — `TAG="${GITHUB_REF_NAME}"`; `RELEASE_URL="https://github.com/${{ github.repository }}/releases/tag/${TAG}"`.
- `release.yml:62-72` — hand-escaped JSON with `\"GLX ${TAG}\"`.
- `.github/CLAUDE.md` "Workflow injection: never interpolate untrusted input into `run:`" rule.

## Recommended fix

```yaml
        run: |
          RELEASE_URL="https://github.com/${GITHUB_REPOSITORY}/releases/tag/${GITHUB_REF_NAME}"
          jq -n --arg tag "$GITHUB_REF_NAME" --arg url "$RELEASE_URL" \
            '{embeds:[{title:("GLX "+$tag),url:$url,description:"A new version of the GLX CLI has been released.",color:5814783,fields:[{name:"Release Notes & Downloads",value:("[View on GitHub]("+$url+")")}]}]}' \
          | curl -sSf -H "Content-Type: application/json" -d @- "$DISCORD_WEBHOOK"
```

`jq --arg` does correct JSON-string escaping; `GITHUB_REPOSITORY`/`GITHUB_REF_NAME` are runner-provided (no `${{ }}` needed) and stay quoted as pure data. `jq` and `curl` are preinstalled on `ubuntu-latest`. Keep the existing `if:`/`env:` keys.

Optional adjacent polish (lower priority): a transient Discord 5xx currently reds an already-successful release because the announce shares the job. If that becomes annoying, move the announce to a separate `needs: release` job (or `continue-on-error: true`). Not filed separately — bundle here if desired.

## Cosign contract safety

No risk. The Discord step is a post-release notification running after the GoReleaser/cosign step; it does not touch the signed artifacts, the OIDC identity, or the workflow file path that the `--certificate-identity-regexp` binds to.

## References

- Security hardening for GitHub Actions (env-staging pattern): https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions
- git-check-ref-format (forbidden ref chars): https://git-scm.com/docs/git-check-ref-format

---
*Found during a deep 2026 best-practice review of `release.yml`.*

noahbjohnson:
## What and why

Four release-pipeline fixes batched into one PR because they touch the same two files (`release.yml`, `.goreleaser.yml`) and share one theme: the release workflow does irreversible, externally-visible work and deserved the hardening its sibling workflows already have.

- **Prereleases are now flagged as prereleases.** `.goreleaser.yml` had no `release:` stanza, so GoReleaser's `prerelease: false` default published every `-beta` tag as a full release — `v0.0.0-beta.1` through `beta.11` are all `prerelease=false` today, and `/releases/latest` resolves to a beta. `release.prerelease: auto` flags `-alpha`/`-beta`/`-rc` tags automatically. (Existing releases are not retroactively re-flagged; that's a one-time `gh api` pass if wanted.)
- **Per-tag concurrency group** (`cancel-in-progress: false`) so the tag-move hotfix flow documented in `.github/CLAUDE.md` — or two pushes in quick succession — cannot interleave asset uploads, cosign signing, and Discord announcements on the same Release. Cancelling is deliberately avoided: killing a run between upload and signing would leave an unsigned release.
- **GoReleaser tool version pinned** to `~> v2.17` (patch-floating within the current minor). The action SHA pin does not pin the tool binary it downloads; `~> v2` resolved to the newest v2.x at run time, so a moved-tag re-run months later could build different artifacts. A comment marks it as Dependabot-invisible so it's bumped deliberately.
- **Discord announce payload built with `jq --arg`** instead of hand-spliced `\"` escapes. The `-beta*` trigger glob admits any non-slash character and git permits `"` in ref names, so a raw `${TAG}` splice could emit malformed JSON. Also replaces inline `${{ github.repository }}` in `run:` with the runner-provided `$GITHUB_REPOSITORY`, per the repo's env-staging rule.

## Related issues

Fixes #1007
Fixes #1008
Fixes #1009
Fixes #1010

## Review focus

The cosign contract is untouched — `signs:` config, `cosign-release` pin, OIDC identity, and workflow path are all unchanged; each issue's "Cosign contract safety" section was checked against the diff. Worth a second pair of eyes on `cancel-in-progress: false` being the right choice (queueing vs cancelling).

## Testing

- `actionlint .github/workflows/release.yml` — clean.
- Smoke-tested the jq payload locally with the adversarial tag `v1.2.3-beta"x`: produces valid JSON with the quote correctly escaped (the old hand-built string breaks on it).
- `prerelease: auto` is GoReleaser-documented behavior; not exercisable without pushing a tag.

## Breaking changes

None for consumers of the artifacts. Operational: once a beta publishes as a true prerelease, automation relying on `/releases/latest` returning a beta will (correctly) start skipping prereleases.


.github/workflows/release.yml
```diff
@@ -78,19 +92,23 @@ jobs:
...
+          RELEASE_URL="https://github.com/${GITHUB_REPOSITORY}/releases/tag/${GITHUB_REF_NAME}"
+          jq -n --arg tag "$GITHUB_REF_NAME" --arg url "$RELEASE_URL" \
+            '{
+              embeds: [{
+                title: ("GLX " + $tag),
+                url: $url,
+                description: "A new version of the GLX CLI has been released.",
+                color: 5814783,
+                fields: [
+                  {name: "Release Notes & Downloads", value: ("[View on GitHub](" + $url + ")")}
```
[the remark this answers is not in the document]
noahbjohnson: Addressed in the follow-up commit. The URL is now built inside jq with the tag passed through @uri, and that single $url binding feeds both the embed url and the Markdown link, so they cannot diverge. Smoke-tested with tag v1.2.3-beta")x: url renders .../v1.2.3-beta%22%29x and the Markdown link stays well-formed; the title keeps the raw tag.

.github/workflows/release.yml
```diff
@@ -13,6 +13,15 @@ on:
...
   contents: read
 
+# Serialize runs per tag: a delete-and-repush of the same tag (the tag-move
+# hotfix flow in .github/CLAUDE.md) or two pushes in quick succession must not
+# race asset uploads, cosign signing, and the Discord announce on one Release.
+# cancel-in-progress MUST stay false — cancelling between asset upload and
+# signing would leave an unsigned/partially-signed release. (#1008)
+concurrency:
+  group: release-${{ github.ref }}
+  cancel-in-progress: false
```
isaacschepp: This group is keyed on a flow the repo no longer allows, so as written it can't serialize the case it's meant to protect.

`.github/CLAUDE.md` on `main` now carries a section titled *"Release pipeline: never move a published tag — cut a new patch tag"* (landed in #1186 after this branch forked), which says the opposite of what this comment cites: delete-and-recreate of a published tag is broken-by-design under GitHub Immutable Releases — recreating the name returns HTTP 422 `tag_name was used by an immutable release` and burns the name permanently. The sanctioned re-run flow cuts a **new** patch tag, which gives a different `github.ref` and therefore a different concurrency group.

Since `on:` here is tag-push only (`v*` / `v*-beta*`), two runs can otherwise never share a ref either — so `release-${{ github.ref }}` is close to a no-op in practice. Keying the group on the workflow makes it cover the one re-run path that is still permitted (a hotfix patch tag pushed minutes behind a still-running release):

```suggestion
# Serialize release runs. The sanctioned re-run flow (.github/CLAUDE.md,
# "never move a published tag") cuts a NEW patch tag rather than moving a
# published one, so a per-ref group would never serialize anything — key on
# the workflow so a hotfix tag pushed minutes behind a still-running release
# queues instead of racing its asset uploads, cosign signing, and Discord
# announce. cancel-in-progress MUST stay false — cancelling between asset
# upload and signing would leave an unsigned/partially-signed release. (#1008)
concurrency:
  group: release
  cancel-in-progress: false
```

Trade-off worth stating explicitly: two unrelated tags pushed at the same moment now queue rather than run in parallel. With `cancel-in-progress: false` that only costs wall-clock on an already rare event, which seems clearly preferable to interleaved signing. If you'd rather keep per-ref grouping, the comment still needs the `.github/CLAUDE.md` reference corrected — citing a deleted, now-forbidden procedure in the release workflow is the kind of thing the next person will follow.

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/release.yml
```diff
@@ -58,7 +67,12 @@ jobs:
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94 # v7.2.3
         with:
-          version: '~> v2'
+          # Pin the GoReleaser tool to a minor (patch-floating), not '~> v2':
+          # the action SHA pin above does not pin the tool binary it downloads,
+          # and a moved-tag re-run months later must not silently build with a
+          # newer GoReleaser whose defaults differ. Dependabot does NOT manage
+          # this input — bump it deliberately. (#1009)
+          version: '~> v2.17'
```
isaacschepp: Two problems with the justification here, and one with the pin itself.

1. **"a moved-tag re-run months later"** rests on the same procedure as the concurrency comment above — the one `.github/CLAUDE.md` on `main` now forbids outright. The pin is still worth having, but the reason is plain reproducibility (the action SHA doesn't pin the tool binary it downloads), not tag surgery.
2. **"the current minor"** isn't accurate: `~> v2.17` resolves to v2.17.1, while GoReleaser v2.18.x shipped before this PR's last update. Please re-check with the command `.github/CLAUDE.md` itself recommends (`gh api repos/goreleaser/goreleaser/tags --jq '.[].name' | head`) and pin whatever is current at merge.
3. **Nothing will tell you when this rots.** The comment correctly notes Dependabot doesn't manage this input, but unlike the gosec pin — which has `.github/workflows/gosec-pin-currency.yml` watching it — there's no currency gate here, so "bump it deliberately" has no trigger. A deliberately-managed pin with no reminder is a pin that stays at v2.17 for a year.

```suggestion
          # Pin the GoReleaser tool to a minor (patch-floating), not '~> v2':
          # the action SHA pin above pins the action, not the tool binary it
          # downloads, so '~> v2' would build a tagged release with whatever
          # v2.x is newest at run time — different artifacts from the same
          # source. Dependabot does NOT manage this input, and no currency gate
          # watches it: bump it deliberately. (#1009)
          version: '~> v2.18'
```

Could the gosec-pin-currency job be generalized to cover this input in a follow-up? Happy for that to be a separate issue rather than scope creep here — but then let's file it before merge so the pin has an owner.

---
_Generated by [Claude Code](https://claude.ai/code)_

.goreleaser.yml
```diff
@@ -39,6 +39,11 @@ archives:
       - CHANGELOG.md
 checksum:
   name_template: 'checksums.txt'
+release:
+  # auto marks -alpha/-beta/-rc tags as GitHub pre-releases, keeping them out
+  # of /releases/latest and the "Latest" badge. Without this, GoReleaser's
+  # default (false) published every beta as a full release (#1007).
+  prerelease: auto
```
isaacschepp, line 46: `prerelease: auto` is the right fix for #1007, but it silently breaks the one install path we document, and that fallout isn't handled in this PR.

`docs/guides/git-workflow.md:222` tells users to install from `/releases/latest/download`:

```bash
base="https://github.com/genealogix/glx/releases/latest/download"
```

`/releases/latest` skips prereleases. Every release that exists today is a beta stored as `prerelease=false`, and this PR deliberately leaves them that way, so after this merges:

- **If the old betas stay as-is** — `/releases/latest` pins to `v0.0.0-beta.11` permanently, because every future beta is now correctly a prerelease and can never take the "Latest" slot. The snippet keeps exiting 0 while installing an ever-staler binary, which is worse than failing: the checksum verification passes and nothing looks wrong.
- **If the betas are retroactively re-flagged** (the `gh api` pass the PR body floats) — there is no non-prerelease release left at all, and the snippet 404s on the first `curl -fsSL`.

Either way the documented path is wrong the moment this lands. The doc already hints at the fix, so promoting it to the default is a two-line change:

```diff
-          # Verify the published checksum before running the binary — never pipe
-          # an unverified download straight into a shell. For full reproducibility,
-          # replace "latest/download" with a pinned "download/vX.Y.Z".
+          # Verify the published checksum before running the binary — never pipe
+          # an unverified download straight into a shell. Pin the version: while
+          # GLX is in prerelease, /releases/latest resolves to nothing usable
+          # (all published releases are betas, which are now flagged as
+          # prereleases and excluded from "latest").
           mkdir -p .bin
-          base="https://github.com/genealogix/glx/releases/latest/download"
+          base="https://github.com/genealogix/glx/releases/download/v0.0.0-beta.11"
```

I'd rather see that ride along here than land as a follow-up, since the window between the two is exactly when someone copies a broken snippet. If you'd prefer to keep this PR to the two workflow files, a tracking issue linked from the PR body works too — but please decide the retroactive re-flag question at the same time, since the doc's correct wording depends on it.

---
_Generated by [Claude Code](https://claude.ai/code)_

isaacschepp (changes requested):
Reviewed the two changed files against `main` (`cb44d31`). The core mechanics hold up — I ran them rather than assuming:

- The jq payload executed locally against `v1.2.3`, `v1.2.3-beta.11`, `v1.2.3-beta"x`, and `v1.2.3-beta.1+meta`: all four produce valid JSON with correct escaping and `@uri` encoding. Piping into `curl -d @-` is safe here (Actions' default shell sets `-o pipefail`, and curl's newline handling on `@-` is harmless for JSON). This is a real fix — the old hand-spliced form does break on the quote case.
- `~> v2.17` is a supported `goreleaser-action` constraint form and resolves to a real release.
- The workflow YAML parses, `concurrency` isn't duplicated (main's `release.yml` has no such block), and `git merge-tree` shows zero conflicts with main.
- The cosign contract is genuinely untouched, as the PR body claims.

Requesting changes on four things — three are comments or config that point at a procedure `main` deleted, one is a user-visible break:

1. **`prerelease: auto` strands the documented install path** (`.goreleaser.yml:46`, inline). `docs/guides/git-workflow.md:222` installs from `/releases/latest/download`; after this lands that either pins to `v0.0.0-beta.11` forever or 404s, depending on whether the old betas get re-flagged. Needs a decision, and ideally the doc fix, before merge.
2. **The concurrency group can't serialize the only re-run path the repo still allows** (`release.yml:16`, inline). Its comment cites the tag-move hotfix flow from `.github/CLAUDE.md`, which #1186 deleted and reversed after this branch forked — that doc now documents delete-and-recreate as broken-by-design under Immutable Releases (HTTP 422, tag name burned permanently).
3. **The GoReleaser pin comment reuses the same forbidden premise and overstates currency** (`release.yml:70`, inline) — `~> v2.17` is not "the current minor" any more, and with Dependabot explicitly not managing the input and no currency gate (unlike `gosec-pin-currency.yml`), it will rot silently.
4. **No `CHANGELOG.md` Unreleased entry.** `CONTRIBUTING.md` step 6 requires one for user-facing changes, and this PR's own "Breaking changes" section describes an externally visible one (consumers of `/releases/latest` start skipping betas). Something like, under `## [Unreleased]` → `### Fixed`:

   > - **Prerelease tags now publish as GitHub prereleases** — `.goreleaser.yml` had no `release:` stanza, so GoReleaser's `prerelease: false` default published every `-beta` tag as a full release and `/releases/latest` resolved to a beta. `-alpha`/`-beta`/`-rc` tags are now flagged automatically; existing releases are not retroactively re-flagged. Also pins the GoReleaser tool version, serializes release runs, and builds the Discord announce payload with `jq --arg`. (#1007, #1008, #1009, #1010)

   Note the changie migration (#858) isn't on `main` — there's no `.changes/` there yet — so this needs a hand-written entry rather than a fragment. If #858 lands first, convert it.

Also: the branch is currently **behind** `main`, so it'll need an update before merge regardless.

Nothing here touches the signing path, and the four fixes themselves are the right calls — it's the surrounding justification that has drifted out from under the branch during its three weeks open.

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/release.yml
isaacschepp: Fixed in 9202e65 — the group is now `release` and the comment cites the current `.github/CLAUDE.md` guidance ("never move a published tag") instead of the deleted flow. `cancel-in-progress: false` is untouched.

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/release.yml
isaacschepp: Fixed in 9202e65. Checked upstream before pinning — `git ls-remote --tags goreleaser/goreleaser 'v2.*'` shows v2.18.1 as the newest stable (v2.19.0 exists only as `-nightly` tags), so the pin is now `~> v2.18` and the comment no longer claims a currency it can't keep. The justification is reproducibility rather than tag surgery.

Filed #1215 for the currency gate, and the comment points at it so the next reader knows the pin is watched-by-issue rather than watched-by-CI. Worth weighing there whether to generalize `gosec-pin-currency.yml` or add a sibling workflow; the check needs to compare minors only, since this pin floats patches deliberately.

---
_Generated by [Claude Code](https://claude.ai/code)_

.goreleaser.yml
isaacschepp: Fixed in 9202e65 — `docs/guides/git-workflow.md` now downloads from `releases/download/v0.0.0-beta.11` with a comment explaining why `latest/download` isn't usable while every release is a prerelease. `make check-links` and markdownlint are clean.

The retroactive re-flag question is now #1216 rather than a decision this PR has to make: the doc reads correctly under either outcome, since it no longer resolves `/releases/latest` at all. That issue also flags checking whether Immutable Releases (#931) permits PATCHing `prerelease` on already-published releases before anyone plans the pass.

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/release.yml
```diff
@@ -78,19 +92,25 @@ jobs:
         if: success() && env.DISCORD_WEBHOOK != ''
         env:
           DISCORD_WEBHOOK: ${{ secrets.DISCORD_RELEASE_WEBHOOK }}
+        # Build the payload with jq --arg (correct JSON-string escaping) instead
+        # of hand-spliced strings: the -beta* trigger glob admits any non-slash
+        # character, and git permits `"` in a ref name, so a raw ${TAG} splice
+        # could produce malformed JSON. The tag is additionally @uri-encoded
```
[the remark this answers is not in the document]
isaacschepp: Good catch — fixed in fcea966. The comment now says "splicing the raw `$GITHUB_REF_NAME` into the body", which is what the step actually does; the surrounding block was re-wrapped to match.

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/release.yml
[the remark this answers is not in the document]
isaacschepp: This one doesn't apply, on two counts.

GitHub Actions does not start a run for a ref *deletion* on an `on: push` trigger — the push webhook fires with `deleted: true`, but workflow dispatch is skipped; deletions are the separate `delete` event, which this workflow doesn't subscribe to. So there is no delete-run to fail early or to queue ahead of anything.

Separately, the flow this assumes is no longer permitted: `.github/CLAUDE.md` (as of #1186) documents delete-and-recreate of a published tag as broken-by-design under Immutable Releases — recreating the name returns HTTP 422 and burns it permanently. The sanctioned re-run cuts a new patch tag. The concurrency group has been re-keyed to the workflow in 9202e65 to serialize *that* flow, which a per-tag group could not.

---
_Generated by [Claude Code](https://claude.ai/code)_

isaacschepp (approved):
Approving. Full disclosure: I pushed the fixes myself rather than handing them back — 9202e65 and fcea966 are mine, so this approval covers @noahbjohnson's original four changes plus my follow-ups, and a second maintainer eye on the latter wouldn't hurt.

What changed since the review:

| Finding | Fix |
|---|---|
| Concurrency group keyed on a deleted flow | `group: release`; comment now cites the current "never move a published tag" guidance. `cancel-in-progress: false` untouched. |
| `~> v2.17` not the current minor; moved-tag premise | `~> v2.18` (verified against upstream: v2.18.1 newest stable, v2.19.0 is nightly-only); comment rewritten around reproducibility and points at #1215. |
| `prerelease: auto` stranding the documented install | `docs/guides/git-workflow.md` downloads from a pinned tag instead of `/releases/latest/download`. |
| Missing changelog entry | Two entries under `## [Unreleased]` → `### Fixed`. |
| Stale `${TAG}` in the announce comment (Copilot's open thread) | Reworded to `$GITHUB_REF_NAME`, block re-wrapped. |

Also merged `main` to clear the behind-base state — that brought in #1181's `go-version-file: go.mod` and the reworded cosign pin comment, no conflicts.

Two follow-ups filed rather than scope-creeping this PR:

- **#1215** — the GoReleaser pin is Dependabot-invisible and has no currency gate; the gosec pin has one, this should too. Worth doing before the pin has a chance to rot.
- **#1216** — whether to retroactively flag the eleven published betas as prereleases. Not a blocker any more (the doc no longer resolves `/releases/latest`), but the repo's "Latest" badge keeps pointing at `v0.0.0-beta.11` until it's decided, and it's worth confirming Immutable Releases (#931) still permits PATCHing that flag.

Verification on my side: YAML parses, `bash scripts/check-links.sh` clean (442 links), `markdownlint-cli2` clean across the 68 linted files, `go build ./...` clean after the merge. CI on `fcea966` is still running — worth a glance before merge, though nothing in this diff touches Go.

The signing path is still untouched end to end: `signs:`, the `cosign-installer` pin, `cosign-release: 'v3.0.6'`, the OIDC identity, and the attestation step are all byte-identical to `main`.

---
_Generated by [Claude Code](https://claude.ai/code)_

isaacschepp:
CI is green across all 39 checks on `fcea966` and the PR is mergeable — approved.

One loose end I'd rather not fix by editing your prose, @noahbjohnson: the description now disagrees with the diff in four places, and since it becomes the merge commit message, the drift would land in history.

- "Per-tag concurrency group" → the group is `release`, keyed on the workflow. A per-tag group couldn't serialize anything, because the flow that would have shared a ref is the delete-and-repush one.
- "the tag-move hotfix flow documented in `.github/CLAUDE.md`" → that doc now documents the opposite (#1186): delete-and-recreate is broken-by-design under Immutable Releases, and the sanctioned re-run cuts a new patch tag.
- "pinned to `~> v2.17` (patch-floating within the current minor)" → now `~> v2.18`; v2.17 stopped being the current minor before this PR's last update.
- "a raw `${TAG}` splice" → the step never defined `TAG`; the comment and the description should both say `$GITHUB_REF_NAME`.

Also worth adding to **Breaking changes**: the documented install snippet in `docs/guides/git-workflow.md` moved off `/releases/latest/download` to a pinned tag, since `prerelease: auto` strands that path while every published release is a beta.

Happy to push the body edit myself if you'd rather not hand-edit it — say the word and I'll mirror the four corrections without touching the rest of your wording.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1204, opened 2026-09-04
fix: Make glx rename rewrite only touched files instead of swapping the archive directory

### LINKED ISSUE #1192: glx rename exits into stale directory in multi-file archive
### Area

cli

### Summary

If I run `glx rename person-old-id person-new-id` in the root directory of a multi-file archive, `glx` exits into the directory where it was executed. But that directory is now empty.

This happens wether I specify `—archive .` or leave that option unspecified.

As expected, the files are now in a new directory with the same name as the original.

But I would expect `glx rename` to exit into that new safely-written directory rather than the now-alarmingly-empty original one.

### Reproduction

In the root directory of a multi-file archive:
1. Run `ls -ia`. Note the listed files and the inode of .
1. Run `glx rename person-old-id person-new-id`
1. Run `ls -ia`. Note the lack of files, and that the inode of . is the same as step 1.

### Context

glx version 0.0.0-beta.11 (bc6f389) 2026-05-28T11:29:22Z
MacBook Pro M1 2021 Tahoe 26.5.2

### LINKED ISSUE #1197: glx rename splits entire archive unexpectedly
### Area

cli

### Summary

Doing a `glx rename` does not  preserve the file organization and instead does a full split.

GOT: It appears to read in every file, do the rename, then write out everything as separate file.

WANT: Read each file independently, do the rename, rewrite the same file if necessary.

### Reproduction

Make single .glx file with two items.  Do a rename on one identifier.  

WANT: same file is preserved.

GOT: Two files.

### Context

- GLX Version:

commit b89f6c4ea5d35cd6ee1a2dd1b84fb2f01dc91f2b (HEAD -> main, origin/main, origin/HEAD)
Author: dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>
Date:   Fri Aug 28 16:49:45 2026 +0000

    chore(deps): bump eslint from 10.8.1 to 10.9.0 in /website (#1190)

noahbjohnson:
## What and why

On a multi-file archive, `glx rename` loaded everything, applied the rename in memory, and wrote a complete fresh serialization through the temp-directory swap shared with `migrate` and `merge`. Two user-visible problems followed:

- The archive directory was replaced by a new inode. A shell sitting in the archive root (or an editor with entity files open) was left on the deleted original, so `ls` showed an empty directory until `cd .` (#1192).
- Every file was re-emitted in the serializer's canonical one-entity-per-file layout and formatting: multi-entity files were split, single-entity files renamed to match their ID, unrelated files picked up indentation/quoting churn, and the standard vocabularies were materialized into `vocabularies/` (#1197). Reproduced on `docs/examples/basic-family`: 6 references updated, 16 files modified, 2 deleted, 14 created.

Rename is a much more targeted operation than migrate or merge, so it now gets its own write path:

- Validate against the whole archive as before (old ID present, new ID free).
- Treat each file as an independent fragment. Files that neither define nor reference the old ID are not touched at all. Files that do are rewritten in place at the same path.
- Multi-entity files keep their grouping. Only a single-entity file whose name is the old ID is moved to the new name, refusing if that path already exists.
- The original bytes of every touched file are held in memory and restored in reverse order if any write fails, so a partial rename cannot leave the archive inconsistent.

Same rename on `basic-family` now touches 7 files, and the directory inode is unchanged. Rewritten files still take the serializer's formatting; byte-exact preservation of untouched lines within a changed file is not attempted.

Adds `RenameInFragment` and `GLXFile.HasEntity` to `go-glx`. `migrate` and `merge` keep the swap, since they legitimately rewrite most of an archive.

## Related issues

Fixes #1192
Fixes #1197

## Review focus

- `planRenameWrites` / `applyFileOps` in `glx/rename_runner.go`: the create-before-delete ordering and the rollback path.
- Whether dropping the temp-dir swap for this one command is the right tradeoff (discussed on #1192 by the reporters, who leaned this way).

## Testing

- New `glx/rename_runner_test.go`: layout and untouched-file preservation (byte-for-byte), multi-entity file keeps its name when one of its entities is renamed, dry run writes nothing, missing/taken ID errors, target-file-exists guard, rollback on write failure (both rewritten and created files), single-file archive path.
- New `RenameInFragment` / `HasEntity` tests in `go-glx/rename_test.go`.
- `go test ./glx ./go-glx` green; `golangci-lint` clean on changed files.
- Smoke-tested against a copy of `glx-archive-westeros` (4943 files): renaming a person touched 6 files, directory inode unchanged, dry run wrote nothing. The demo archive's 688 pre-existing validation errors are identical before and after.

## Breaking changes

None.


isaacschepp (dismissed):

isaacschepp (approved):
Approving. Reviewed the whole diff against `main` — the surgical write path, the go-glx rename primitives, and the new e2e harness. The core design (validate against the merged archive, treat each file as an independent fragment, plan then preflight then write, in-memory rollback) is sound, and the earlier review rounds have already covered symlinks, symlinked parents, case-folded collisions, per-type scoping, mode preservation, and the file-changed guard. `go test ./...` is green.

I found four issues and pushed a fix for them in 0b3a12c:

1. **`--dry-run` skipped the filesystem checks entirely.** It returned before `applyFileOps`, so `preflightFileOps` never ran. On an archive where a touched `.glx` file is a symlink, `glx rename person-a person-b --dry-run` printed `Updated 2 reference(s) in 2 file(s)` and exited 0, while the identical real run exited 1 with `refusing to rewrite a symlinked archive file`. Same divergence for `ErrRenameFileChanged` and for a move target occupied by something the loader did not collect — precisely the questions `--dry-run` exists to answer.

2. **The success summary was printed before the plan was validated.** A run that then failed still wrote `Renaming X → Y` / `Updated N …` to stdout before the error reached stderr.

Both are fixed by ordering and preflighting the plan up front and sharing that result between the dry run and the write. `applyFileOps` splits into `orderFileOps` + `executeFileOps` so the checked, ordered slice is the one that gets executed rather than being rebuilt.

3. **`RenameResult.RefsUpdated` doc contradicted its value.** The field says "number of reference fields updated", but it is now references *plus one* for the moved map key, and the CLI printed it as "reference(s)". Renaming Robert in `basic-family` updates 5 reference fields and reported 6. The field comment now matches the code and both CLI messages say `change(s)`; the e2e assertion is updated.

4. **Preflight accepted a plan targeting the same path twice** (open thread r3981904746). Not reachable through the current planner, but nothing enforced the invariant and rollback could not restore the state between two writes to one file. Added an up-front duplicate-path check with a new `ErrRenameDuplicatePlanPath`.

New tests cover the dry-run preflight, the no-output-on-refusal case, and the duplicate-target guard; the CHANGELOG entry is extended.

One thing I looked at and deliberately did not change: the TOCTOU window between preflight and execute (thread r3982221996). Closing it properly means writing through an `os.Root`, which is a larger change than this PR should carry, and the realistic attacker already needs write access to the archive directory. Worth a follow-up issue rather than a blocker.

I also confirmed the two behaviours the PR promises that most deserve checking: unrelated files come back byte-identical, and rewritten files lose unknown top-level keys and comments — the latter is unchanged from the previous full-rewrite path (which lost them from *every* file), so this is strictly an improvement, as the PR body says.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1218, opened 2026-09-10
fix: Skip temporal properties in the analyze conflict check

### LINKED ISSUE #1213: `glx analyze` conflict check ignores assertion dates, so temporal properties are always flagged
### Area

cli

### Summary

`glx analyze --check conflicts` groups assertions by person and property but never reads their `date`, so any property that legitimately changes over a life is reported as a conflict. Someone who moved house twice gets a HIGH conflict on `residence`; a woman recorded under her maiden and married names gets one on `name`.

The shipped `docs/examples/temporal-properties/` archive fails the check, unmodified.

### Reproduction

```bash
glx analyze --archive docs/examples/temporal-properties --check conflicts
```

Observed:

```
=== Research Gap Analysis: 1 issues found ===

CONFLICTS (1)
  HIGH person-john-smith-1850         John Smith — residence has 2 conflicting values: Leeds [high], London [high]
```

Expected: no conflict. The two residences carry `date: "FROM 1850 TO 1870"` and `date: "FROM 1870 TO 1920"`, which is what `specification/4-entity-types/assertion.md` prescribes for the `date` field: "Each assertion can target a specific temporal entry using `date`."

### Context

- The grouping key is `{personID, property}` (`glx/analyze_conflicts.go:28-31`, used at `:53`). Neither `date` nor `status` is read, so an assertion already resolved as `status: disputed` is reported at the same severity as an ordinary move.
- #156, which added the check, motivates it with `born_at`, a non-temporal property.
- `glx evidence <person> residence` behaves the same way, calling two residences 24 years apart an "inconclusive tie".
- Reproduced identically on `v0.0.0-beta.11` and on `main` (b7bb6ac). macOS 26.6.2 arm64, go1.27.1.
- On an 811-file personal archive the check yields 12 findings, of which 11 are dated temporal values.

MoSattler:
## What and why

Issue #1213. `glx analyze --check conflicts` flags John Smith in the temporal-properties example because he lived in Leeds in 1851 and in London from 1870. The check compares assertion values and ignores their date, so any move looks like a contradiction.

This change leaves out every property the vocabulary flags `temporal: true`. Residence, occupation and name no longer produce conflicts. Non-temporal properties such as `buried_at` are checked as before.

The gap: two residences with overlapping dates are also not checked. Once #1199 lands, the check could parse both dates and report only the pairs whose ranges overlap.

## Related issues

Fixes #1213

## Review focus

Skip versus group by date.

## Breaking changes

Same-date contradictions on temporal properties are no longer reported.


MoSattler:
There is a second version of this fix in #1227. It keeps temporal properties in the check and compares the dates, so two residences in the same year are still reported, with the date next to each value. I opened it separately because I do not know which of the two you prefer; merge one and close the other.

This branch is rebased on main, the code is unchanged.

noahbjohnson:
Closing in favor of #1227

PR #1227, opened 2026-09-11
fix: Compare dates on temporal properties in the analyze conflict check

### LINKED ISSUE #1213: `glx analyze` conflict check ignores assertion dates, so temporal properties are always flagged

MoSattler:
## What and why

Issue #1213. The conflict check compares assertion values and ignores their date, so a residence in Leeds in 1851 and one in London from 1870 read as a contradiction.

For a property the vocabulary flags `temporal: true`, two values now conflict only when their dates overlap. A year covers the whole year, `BET` and `FROM … TO` the range, a lone `FROM` or `TO` an open end, a missing date all time. `ABT 1851` or `AFT 1851` counts as 1851. Non-temporal properties such as `buried_at` are checked as before.

The span arithmetic sits in `analyze_conflicts.go` because `glxdate` has no range comparison yet.

#1218 fixes the same issue by skipping temporal properties altogether.

## Related issues

Fixes #1213

## Review focus

A missing date counts as all time, so an undated entry stays in the report. Leaving it out would hide a married name recorded without a date next to the birth name.

## Breaking changes

None.


noahbjohnson (changes requested):
Right direction, but two bugs in `spanOf` make `analyze` miss conflicts that `main` catches. Built this branch and `eed00ad9` and ran both:

|                               | `main`   | PR              |
| ----------------------------- | -------- | --------------- |
| 50 years apart                | conflict | none ✅ the fix |
| `AFT 1900` vs `1950`          | conflict | none ❌         |
| `BEF 1900` vs `1880`          | conflict | none ❌         |
| `FROM 1920 TO 1870` vs `1900` | conflict | none ❌         |

**`BEF`/`AFT` collapse to the year they exclude.** Both give `[19000101, 19001231]` for 1900, so `AFT 1900` never overlaps 1950. Note `extractRelationshipBoundaryYear` (`analyze_suggestions.go:304`) already implements the opposite rule, in the same command.

**An inverted span overlaps nothing.** `FROM 1920 TO 1870` is valid and validate is silent on it, but `spanOf` returns `lo > hi` and the entry drops out of every comparison. The `Year() == 0` guard only covers the start, so a range end with no year does the same; validate only warns on those and analyze runs regardless.

Both are in `spanOf`: open `BEF`/`AFT` to `-Inf`/`+Inf` on the far side, guard each range endpoint for a missing year, and treat `lo > hi` as all-time. I tried that locally and it's green with the branch's tests.

Two more:

- **Single-file archives don't get the fix.** `loadArchiveForAnalyze` skips `mergeStandardVocabularies`, so `--archive docs/examples/temporal-properties` is clean but `--archive .../archive.glx` still prints the #1213 conflict. One call, same as `evidence_runner.go:124`.
- **The temporal gate is untested.** `isTemporalProperty` → `return true` passes the whole `./glx` suite. A non-temporal property that still conflicts regardless of dates would cover it.

PR #1259, opened 2026-09-18
docs: Make ADR-0006 the source of truth for the no-I/O rule

noahbjohnson:
## What and why

`docs/decisions/0006-go-glx-library-pure.md` stated its own decision by quoting `go-glx/CLAUDE.md` and linking to it ("The `go-glx` package MUST NOT perform filesystem I/O. From [`go-glx/CLAUDE.md`](…):"). That inverted the relationship between the two files: an ADR is the project's decision record, while a `CLAUDE.md` is an agent guide that is not reviewed as policy, is excluded from the published site, and can be edited without a decision being made. The ADR is also published on the website, so the link put an agent guide in front of public readers as the authority for a public architectural rule.

This PR makes the ADR state the rule directly. In doing so it carries over the substance the ADR's quote had elided behind an ellipsis:

- the full prohibition list — `os.MkdirAll`, `os.Stat`, `os.ReadDir`, and `filepath.Join` used to build a path for a file operation, none of which the ADR previously named;
- the complete allowed set — `io/fs.FS` alongside `io.Reader`, `io.Writer` and `[]byte`, plus `go:embed` for fixture data the library itself needs.

`go-glx/CLAUDE.md` now names the ADR as the source of truth and keeps only the quick-reference summary, with a note that nothing in it may contradict the ADR.

A repo-wide grep of `docs/`, `specification/`, `README.md`, `CONTRIBUTING.md`, `SECURITY.md`, `SECURITY-POSTURE.md` and the package READMEs finds no remaining link to any `CLAUDE.md`; this ADR line was the only one.

Related: the general rule this follows — a `CLAUDE.md` is never the source of truth, and public docs must never link to one — is being added to the root `CLAUDE.md` in #1252. This PR is the second instance of that cleanup and is independent of it.

## Related issues

None — no issue was filed for this.

## Review focus

- Whether the rule as now stated in the ADR is complete and correct, since it is no longer backed by a quote. The wording was taken from `go-glx/CLAUDE.md` rather than invented, but the two are now separate texts and only the ADR is authoritative.
- The link from `go-glx/CLAUDE.md` to the ADR is a relative `.md` path, which is the repo convention and is safe here (a `CLAUDE.md` is not published to the site, so nothing rewrites the href).

## Testing

- `make check-links` — pass (727 relative links)
- `make check-memory-drift` — pass (validates the new relative path in `go-glx/CLAUDE.md` resolves)
- `npx markdownlint-cli2` — 0 issues
- `npx vitepress build` — pass (the ADR is a published page)
- `make test` skipped: no Go code changed.

## Breaking changes

None.


noahbjohnson:
Duplicate of #1258, which carries the more complete rewrite.

Issue #91, opened 2026-03-09
feat: read-only web viewer for sharing archives


isaacschepp:
## Summary

A `glx view` command that exports a GLX archive as a static HTML site — a browsable family tree with timelines, person profiles, and source links. Designed for sharing with non-technical family members who won't install CLI tools or read YAML.

## Motivation

GLX archives are researcher-facing: YAML files, command-line tools, evidence chains. But the end audience for genealogy research is often family — grandparents, cousins, aunts — who want to browse the tree, read narratives, and see photos. A static HTML export bridges that gap without requiring a server, database, or account.

## Proposed CLI interface

```
# Generate static site in ./site/
glx view --archive . --output ./site

# Serve locally for preview
glx view --archive . --serve
# → Serving at http://localhost:8080

# Export with embedded media (self-contained)
glx view --archive . --output ./site --embed-media
```

### Person profile page
- Name, vital dates, portrait photo (if media exists)
- Life timeline: birth → census appearances → marriages → military → death → burial
- Family section: parents, spouses, children (linked to their profiles)
- Source citations with expandable footnotes
- Research notes (if present)
- Associated media gallery (photos, documents)

### Family tree view
- Pedigree chart (ancestors) — collapsible/expandable
- Descendancy chart — from a selected ancestor
- Interactive: click any person to navigate to their profile

### Timeline view
- Chronological view of all events for a person or family group
- Events plotted on a visual timeline with dates and locations

### Source index
- List of all sources in the archive
- Each source links to the assertions and persons it supports
- External links (catalog URLs, FamilySearch ARKs) preserved as clickable links

### Place index
- List of all places referenced in the archive
- Each place links to events and persons associated with it
- Optional: embedded map view if coordinates are present

### Search
- Client-side search (no server needed) using a pre-built JSON index
- Search by name, place, date range, or source title

### Static site generation
- Go templates or embedded HTML/CSS/JS in the `glx` binary
- Single `glx view` command generates all HTML, CSS, JS, and media into an output directory
- No build tools, no npm, no external dependencies for the user
- Output is a self-contained directory that can be opened with `file://` or hosted anywhere (GitHub Pages, S3, Netlify)

### Styling
- Clean, readable typography — optimized for narrative text
- Responsive layout for desktop and mobile
- Light/dark mode toggle
- Print-friendly CSS for person profiles

### Media handling
- Default: reference media files with relative paths (user copies media alongside HTML)
[...]

PR #1266, opened 2026-09-18
feat: Plot the published life timeline to scale

isaacschepp:
## What and why

Second of the three follow-ups #91 still tracks for the static site. `glx publish` already listed a person's events in order, but every row had the same height, so a decade with no records looked exactly like a decade with one event a year.

Each profile's **Timeline** section now opens with a proportional strip:

- a bar spanning the person's first to last dated event,
- one dot per dated event, placed by year,
- an axis labelled in round years.

Same approach as the family charts in #1265: inline SVG generated in Go at publish time — no JavaScript, no chart library, no network request — scaled to the page width by the stylesheet and themed from the site's CSS custom properties. Each dot carries the full event as an SVG `<title>`, so hovering names the date, event and place. The chronological list is untouched below it: the strip is an overview, not a replacement, and undated events still appear only in the list.

Details worth knowing:

- Axis labels cannot collide. The step is the smallest of 1/2/5/10/20/25/50/100/…/2000 years that keeps the label count within six; the span's endpoints are always labelled; an interior label closer than 40 user units to its neighbor is dropped rather than drawn overlapping.
- Years before the common era render as `44 BCE`, matching the rest of the site — `eventYear` now shares that helper instead of repeating the formatting.
- `timelineRow` gains the event's parsed year, which is what the strip plots.
- A person whose dated events fall in a single year — or who has none — gets no strip, because a proportional view of one point says nothing the list does not.

## Related issues

Refs #91. Not a closing keyword: the map view, the third follow-up, comes next.

**Stacked on #1265** (pedigree/descendancy charts) — this PR targets that branch, since both touch `person.html.tmpl` and `style.css`. Retarget to `main` once #1265 merges; review only the commit on this branch.

## Review focus

- Tick selection in `stripTicks`/`stripTickYears` — label budget, endpoint handling, the minimum-gap rule.
- Whether the strip belongs above the list (current) or beside the heading.
- The "one distinct year ⇒ no strip" cutoff.

## Testing

- `go test ./glx/...` — new `glx/publish_timeline_test.go` covers the nil cases (no rows, undated rows, single year), proportional dot placement including both endpoints, tick labels and their minimum gap, the label budget on a 2000-year span with a BCE endpoint, tooltip content, the parsed year on timeline rows, and the rendered markup. Two pre-existing Windows-only failures (`TestSafeWriteMultiFileArchive/refuses_to_wipe_stale_backup_with_unrecovered_media_binaries`, `TestPreflightFileOps_ReportsUnexpectedStatError`) reproduce unchanged on `main`.
- `golangci-lint run ./glx/...` — clean for the changed files.
- Manual: published `docs/examples/complete-family` and the 31-person Shakespeare import, then checked both themes in Chrome. The Shakespeare strip (1531–1646) shows the 1560s cluster of births and the sparse later decades, and the axis drops `1640` because it would have collided with the `1646` endpoint.

## Breaking changes

None. Additive to the generated output.


noahbjohnson (changes requested):
Needs conflicts resolved

Issue #1211, opened 2026-09-08
glx analyze: vital_record sources cited on event-subject assertions are not counted


MoSattler:
### Area

cli

### Summary

`glx analyze` says a person has no vital record source when the archive has one, cited on their
birth event.

I might be holding it wrong, and the right answer may be that I should model this differently. But
two of the example archives in this repo produce it, so I thought I'd ask.

### Reproduction

```bash
git clone https://github.com/genealogix/glx.git
cd glx/docs/examples/assertion-workflow
glx analyze
```

```text
→   person-alice-chen    Alice Chen — search vital records (dates exist but no vital record source)
→   person-robert-chen   Robert Chen — search vital records (dates exist but no vital record source)
```

Both have one. Robert's birth date comes from `source-nyc-birth-records` (`type: vital_record`)
through `assertion-robert-birth`, subject `event-robert-birth`. Alice's is
`source-mass-birth-records`, same shape.

`docs/examples/basic-family` does it too: two `vital_record` sources, all four people flagged.

MoSattler:
I dug into this a bit: the check only counts evidence attached straight to a person, so a certificate cited on someone's birth event never registers as theirs. I'm happy to take a swing at counting event-cited evidence for the people the event is about, skipping witnesses and informants. A few of the other `analyze` checks make the same assumption, so tell me if you want those too or just vital records.

Issue #186, opened 2026-03-16
Reduce US-centric defaults in CLI commands


noahbjohnson:
Several CLI commands have US-centric assumptions baked into their defaults:

- **`census add`** (PR #104): Auto-generated source title defaults to "U.S. Federal Census"; `type` field only has special handling for "federal" and "state" (American concepts)
- **`coverage`**: Suggests US census years and US-specific vital record types
- **`analyze`**: Census year suggestions assume US decennial census schedule

The template/input formats themselves are flexible enough for international use — the bias is in the defaults and suggestions, not the spec or data model. A UK user doing the 1841 census can override `source.title`, but the ergonomics aren't great.

### Possible approaches
- Make defaults locale/country-aware (heavier lift)
- Drop country-specific prefixes from defaults (e.g., "Federal Census" instead of "U.S. Federal Census")
- Add a `country` or `locale` field to templates/config that adjusts suggestions
- Document workarounds for non-US users in the meantime

### Scope
Code-level defaults only — the spec and schema are not affected.

isaacschepp:
A repro with **no US connection at all**, which is a simpler case than the immigrant scenario in #261 — this person was born, married and died in one Mecklenburg-Strelitz village and never left.

Single-person archive on `main` @ `06a4dcb`. Every place is parented under a `place-hrr` / `place-mecklenburg-strelitz` chain, the repository carries `country: Germany`, and the death event's place is `place-liepen`:

```text
$ glx coverage person-hollnagel-michael-david
Source Coverage for Michael David Hollnagel
Born: BET 1737 AND 1747 | Died: 1807-08-31, Liepen

  Census Records:
    [ ] 1790 US Census (age ~53)
    [ ] 1800 US Census (age ~63)

$ glx analyze
SUGGESTIONS (3)
  → ... — search 1790 census (alive, no census event)
  → ... — search 1800 census (alive, no census event)
  → ... — search vital records (dates exist but no vital record source)
```

Two details that may help whoever picks this up:

**The year list is duplicated.** There are two independent hard-coded slices, so a fix has to touch both or they will drift:

- `glx/analyze_suggestions.go:83` — `usFederalCensusYears` (1790–1950)
- `glx/coverage_runner.go:28` — `usCensusYears` (1790–1950)

[...]

PR #1278, opened 2026-09-18
fix: require evidence before glx coverage counts a record as found

### LINKED ISSUE #1267: glx coverage: an uncited event counts as a found vital record, inflating the coverage score
### Area

cli

### Summary

`glx coverage` marks a vital record as **found** when the archive holds only an *event* of that
type, with no citation and no source behind it. The event alone flips the checkbox and counts
toward the coverage percentage.

In `glx/coverage_runner.go`:

```go
// Birth record
birthFound := hasEventType(events, glxlib.EventTypeBirth) || hasSourceType(sources, glxlib.SourceTypeVitalRecord, "birth")
// Death record
deathFound := hasEventType(events, glxlib.EventTypeDeath) || hasSourceType(sources, glxlib.SourceTypeVitalRecord, "death")
```

The left side of each `||` is satisfied by existence alone. `collectPersonEvents` gathers events
purely by participation — it never looks at assertions, citations or sources — and `hasEventType`
only checks that the type is present.

The effect is backwards for an evidence-first tool: a **conclusion** is counted as a **record**,
and the score is inflated for exactly the people whose records are missing, which are the ones
coverage is most useful for.

### Reproduction

An archive with one person, a death entry from a parish register, and a birth event that is an
explicit estimate reckoned back from the age band in that death entry. No birth record exists —
the parish demonstrably wrote no baptisms in the relevant years.

```bash
glx init repro && cd repro
glx add person --id person-x --given "Michael David" --surname "Hollnagel" --sex male
glx add place  --id place-liepen --name Liepen --type locality
glx add event  --id event-death --type death --date 1807-08-31 --place place-liepen --principal person-x
# birth is an estimate, not a record: no place, no source, derived from the death entry
glx add event  --id event-birth --type birth --date "BET 1737 AND 1747" --principal person-x
glx coverage person-x
```

```text
  Vital Records:
    [x] Birth record (via event-birth)
    [x] Death record (via event-death)

  Coverage: 3 of 8 expected records found (37%)
```

Both are reported found. The birth record does not exist and never did.

The `SourceRef` does disclose the provenance (`via event-birth`), so the information is on screen —
but `Found` is `true` and the percentage counts it, which is what a caller reads programmatically
via the JSON output (`"found": true`).

### Context

Suggested shape of a fix, in rough order of preference:

1. Require the event to carry supporting evidence — an assertion whose subject is the event and
   which has at least one citation or source — before treating it as a found record. The
   `citation -> source` walk already exists in `glx/analyze_suggestions.go`
   (`suggestVitalRecords`) and could be lifted to a shared helper.
2. Failing that, split the checklist into *record found* and *event recorded (unsourced)* so the
   percentage only counts the former.

A related wrinkle, which may be the same fix or a separate one: an event can be evidenced by a
source whose type is not `vital_record` — a `church_register` is the vital record for most of
Europe before civil registration. Restricting the right-hand side of the `||` to
`SourceTypeVitalRecord` will keep under-counting those archives even after the event-existence
hole is closed. That overlaps #186 (US-centric defaults) and #1211 (event-subject assertions
skipped), so it may be better handled there.

Found while building a single-person archive from parish-register evidence on `main` @ `06a4dcb`.

isaacschepp:
Closes #1267.

`glx coverage` marked a record **found** when the archive held only an *event* of that type, with nothing citing it. The left side of each `||` was satisfied by existence alone:

```go
birthFound := hasEventType(events, glxlib.EventTypeBirth) || hasSourceType(sources, glxlib.SourceTypeVitalRecord, "birth")
```

`collectPersonEvents` gathers events purely by participation, so a **conclusion** counted as a **record** — and the score was inflated for exactly the people whose records are missing, which are the ones coverage is most useful for.

## The rule now

A record counts as found only when evidence backs it:

- a source about the person, as before, or
- an event that is the subject of an assertion resolving at least one citation or source.

Events carry no `sources` or `citations` fields of their own — under the GLX evidence model the event is the conclusion and the assertion pointing at it is the support — so that walk is the whole question. `eventsWithEvidence()` builds the index once per run; `collectPersonEvents` stamps each event with it.

A **dangling** citation or source reference does not count. Reference validation already reports it as an error, and honoring it here would let a typo stand in for a record.

## Scope: the same hole existed in five more places

The issue reports birth and death, but `hasEventType(...) ⇒ found` governed the whole checklist. All are corrected, since leaving them would keep the same inflation under a different label:

| Record | Counted on |
|---|---|
| Birth, Death | a birth/death event existing |
| Marriage | a marriage event existing |
| Probate/will | a probate or will event existing |
| Church | a baptism or christening event existing |
| Federal census | a census event with a matching year |
| State census | a census event matching year + state |

## An unevidenced event is reported, not dropped

Option 2 in the issue — split *record found* from *event recorded (unsourced)* — is folded into option 1 rather than traded against it. The record is not found and does not count, but the event is named:

```text
  Vital Records:
    [ ] Birth record -- HIGH PRIORITY -- event-birth is recorded, but no citation or source backs it
    [ ] Death record -- event-death is recorded, but no citation or source backs it
```

So a conclusion recorded without a source stays distinguishable from a record nobody has looked for, on screen and in `--json`.

## The issue's repro

```console
$ glx coverage person-x            # before
    [x] Birth record (via event-birth)
    [x] Death record (via event-death)
  Coverage: 3 of 8 expected records found (37%)

$ glx coverage person-x            # after
    [ ] Birth record -- HIGH PRIORITY -- event-birth is recorded, but no citation or source backs it
    [ ] Death record -- event-death is recorded, but no citation or source backs it
  Coverage: 0 of 8 expected records found (0%)
```

And the positive case still counts — `docs/examples/assertion-workflow`, whose birth is cited on an assertion whose subject is the birth event (the shape #1211 describes):

```console
$ glx coverage --archive docs/examples/assertion-workflow person-alice-chen
    [x] Birth record (via event-alice-birth)
```

## `glx proof` inherits the correction

`collectProofGaps` draws its evidence gaps from this checklist, so the fix reaches it. Previously it could assert a birth record was present in the same report that said the birth was not established:

```diff
+    [ ] Birth record -- HIGH PRIORITY -- event-birth-jane is recorded, but no citation or source backs it
-    Birth not yet established. Priority: locate 1880 US Census; 1890 US Census; 1900 US Census.
+    Birth not yet established. Priority: locate 1880 US Census; 1890 US Census; 1900 US Census; Birth record.
```

## Expect scores to fall

The old numbers were inflated, so correcting them lowers them. Across the example archives:

| Archive | Person | Before | After |
|---|---|---|---|
| assertion-workflow | person-alice-chen | 1 of 6 | 1 of 6 |
| basic-family | person-robert-thompson | 2 of 18 | 1 of 18 |
| complete-family | person-john-smith-1850 | 3 of 15 | 2 of 15 |
| complete-family | person-jane-smith-1876 | 2 of 14 | 0 of 14 |
| single-file | person-a1b2c3d4 | 3 of 14 | 1 of 14 |

Spot-checked rather than assumed: `complete-family` has exactly one event-subject assertion with citations (`assertion-john-birth-date` → `event-birth-john`), and that is exactly the one record that stays found for John. Jane's birth and death events have no assertion at all.

## Two adjacent nits, both on lines this change touches

- **Nondeterminism.** `collectPersonEvents` and the marriage-event fallback walked `archive.Events` in map order, so a person with two events of one type could report a different one run to run. Both now walk sorted IDs, which also makes the new refs in the notes stable.
- **`source_ref` on a not-found record.** It was populated unconditionally from the event; the text output hid it but JSON did not, and a caller reads a `source_ref` as evidence. It is now set only when the record is found — and when the find came from a source rather than an event, it names that source instead of being empty.

## What this deliberately does not fix

The source side of the check still recognizes only `SourceTypeVitalRecord`, so an archive whose births are evidenced by a `church_register` — the vital record for most of Europe before civil registration — stays under-counted. The issue calls that out as overlapping #186 and #1211, and it is left to them. Note the event path is type-blind by construction, so a birth event cited to a parish register does now count; what remains is the person-subject source path.

## Verification

```console
$ go test ./glx/...                             # coverage suite green
$ golangci-lint run --new-from-rev=origin/main  # 0 issues (CI uses only-new-issues)
$ make tidy-check check-code-drift check-memory-drift validate-examples   # all clean
$ npx markdownlint-cli2 ...                     # 0 issues
$ ./bin/glx docs --output ./docs/cli/           # help text regenerated
```

`go test ./...` has the two pre-existing Windows-only failures (`TestSafeWriteMultiFileArchive`, `TestPreflightFileOps_ReportsUnexpectedStatError`). Both fail identically on unmodified `main`, verified by stashing this branch; CI runs on `ubuntu-latest`, where they pass.

Tests added: the issue's exact archive as a fixture, evidence arriving via citation and via source, the score moving by one when the birth event is cited, the map-order case, dangling references, person-subject assertions evidencing no event, the census and marriage variants, and the `--json` shape a programmatic caller reads.


isaacschepp:
Superseded by #1303 — same change, resolved against current `main` as one commit. Closing this in favour of that.

Issue #1120, opened 2026-07-12
GEDCOM import: embedded citations create one synthetic source per occurrence — no dedup floods archives with identical placeholder sources


noahbjohnson:
## Problem

`createSyntheticSourceFromEmbeddedCitation` (`go-glx/gedcom_evidence.go:~395`) mints a fresh source ID and entity for **every** embedded inline `SOUR` citation — there is no deduplication, so N occurrences produce N sources. When the inline `SOUR` carries no value and no `TEXT` subrecord, the title falls back to the generic `Embedded Citation (No Source Description)` — the shape Ancestry-style exports produce on nearly every fact.

Real-world impact: a single GEDCOM import produced an archive with **2,094 sources, ~2,000 of them identical** "Embedded Citation (No Source Description)" placeholders (each carrying the "Source created from embedded GEDCOM citation" note) — burying the ~90 real sources and attaching a placeholder evidence ref to nearly every event.

## Current behavior

- Per occurrence: `sourceID := generateSourceID(conv)` + `conv.GLX.Sources[sourceID] = source`, with no lookup of previously created synthetics.
- Title precedence: `SOUR` value (with CONT/CONC) → `TEXT` subrecord → generic fallback.

## Proposal

Dedupe synthetic sources by content identity within a conversion run — key on (title, `properties.description`, notes) after extraction, and reuse the first created source id for subsequent identical embedded citations. At minimum, collapse the no-description generic fallback into one shared placeholder per import. Keep `Stats.SourcesCreated` counting actually-created entities so import summaries stay honest.

## Notes

- Promoting embedded citations to SOURCE_RECORDs per the GEDCOM spec recommendation is correct behavior — the defect is only the missing dedup.
- Related but distinct: #494 (SOUR.TEXT vs SOUR.NOTE mapping), #515 (QUAY → confidence mapping), #96 (platform-specific import adapters).
- Found via a downstream Oracynth import; UI screenshots of the flooded source list available on request.

PR #1282, opened 2026-09-18
fix: Export SOUR for event-subject assertions

### LINKED ISSUE #1269: GEDCOM export drops SOUR for event-subject assertions, silently losing the evidence chain for BIRT/DEAT/MARR
### Area

import-export

### Summary

GEDCOM export drops the evidence for every **event-subject** assertion. A birth, death or marriage
whose date and place are properly cited through `--subject-event` assertions exports with no `SOUR`
subrecord at all, while a person-subject property on the same individual, citing the same citation,
exports its `SOUR` correctly.

This is silent loss on the interchange path: the file is valid GEDCOM, the facts are all present,
and only the evidence chain is missing — so nothing downstream can tell that the dates were ever
sourced.

`go-glx/gedcom_export_person.go:74`, `buildPersonPropertyAssertionsIndex`, is the only index that
feeds `exportAssertionSourceRefs`:

```go
for _, assertion := range expCtx.GLX.Assertions {
    personID := assertion.Subject.Person
    if personID == "" || assertion.Property == "" {
        continue
    }
```

`assertion.Subject.Event` is never read anywhere in the exporter — the one other reference to it in
`go-glx/gedcom*.go` is in a test. The doc comment on the function is candid about the scope
("SOUR on NAME, OCCU, RESI, and other records"); event records are simply not in it.

### Reproduction

Single-person archive on `main` @ `06a4dcb`. The death date, the death place and the occupation all
cite the same citation, `citation-death-1807`. The difference is only the subject: the first two are
`--subject-event`, the third is `--subject-person`.

```bash
glx add assertion --subject-event  event-death-1807 --property date  --value 1807-08-31 \
  --citation citation-death-1807 --confidence high
glx add assertion --subject-event  event-death-1807 --property place --value place-liepen \
  --citation citation-death-1807 --confidence high
glx add assertion --subject-person person-x        --property occupation --value "Arbeitsmann" \
  --citation citation-death-1807 --confidence high

glx export . -f 70 -o out.ged
```

```text
1 DEAT
2 DATE 31 AUG 1807
2 PLAC Liepen, Mecklenburg-Strelitz, Heiliges Römisches Reich
1 OCCU Arbeitsmann (day labourer)
2 SOUR @S1@
3 PAGE Liepen death register, page 200, entry of 31 August 1807
3 NOTE The age column reads 60 = 70, a decadal band, not a single figure. …
```

`OCCU` gets `SOUR @S1@` with its `PAGE` and `NOTE`. `DEAT` gets nothing, though its date and place
are cited to that same page. `BIRT` likewise exports bare, though its date carries two citations.

Same result with `-f 551`.

### Context

This matters more than the neighbouring event-subject issues because it is lossy rather than
merely under-reported. #1211 (`analyze` misses event-subject vital records), #713 (`stats` entity
coverage) and #1267 (`coverage` counts uncited events) all leave the archive intact and get the
reporting wrong. Here the evidence does not survive the round trip out of GLX, and GEDCOM is the
format people will hand to someone who does not run GLX.

Two things to decide, which is why I have not sent a PR:

1. **Where the `SOUR` belongs.** GEDCOM 7 allows `SOUR` under an event structure, so the natural
   mapping is the event's own record (`1 DEAT` / `2 SOUR`) rather than the `INDI`. That means
   indexing event-subject assertions by `(eventID, property)` and calling
   `exportAssertionSourceRefs` while the event record is being built, not from the person index.
2. **Whether `place` and `date` assertions both map to the same event `SOUR`.** Two assertions
   citing one citation should probably not emit `SOUR @S1@` twice under one `DEAT`; some dedup by
   `(sourceXRef, page)` is wanted, in the same spirit as #1120 on the import side.

**The repo's own example reproduces it**, which makes a regression test easy:

```bash
cd docs/examples/complete-family && glx export . -f 70 -o out.ged
```

`assertions/assertion-john-birth-date.glx` has `subject.event: event-birth-john` with two
citations (`citation-john-birth`, `citation-census-john`), and `assertion-john-birth-place.glx`,
`assertion-marriage-date.glx` and the rest are the same shape. Every `BIRT` and `DEAT` in the
resulting file exports bare:

```text
0 @I2@ INDI
1 NAME John /Smith/
1 BIRT
2 DATE 15 JAN 1850
1 DEAT
2 DATE 20 JUN 1920
```

No `SOUR` anywhere under either, for any of the three people.

Found while building a single-person archive from parish-register evidence.

isaacschepp:
## What and why

Before: a birth, death or marriage whose date and place were cited through assertions with `subject.event` exported with no `SOUR` subrecord at all. The file was valid GEDCOM with every fact present and only the evidence chain missing, so nothing downstream could tell the dates had ever been sourced. A person-subject property on the same individual, citing the same citation, exported its `SOUR` correctly — the difference was only the subject. Exporting the repository's own `docs/examples/complete-family` reproduced it: every `BIRT`, `DEAT` and `MARR` came out bare.

After: the evidence rides with the event.

```text
1 BIRT
2 DATE 15 JAN 1850
2 PLAC Leeds, Yorkshire, England
2 SOUR @S2@
3 PAGE Entry 145, Page 23, January 1850
3 NOTE Primary source - parish baptism register
2 SOUR @S1@
3 PAGE District: Leeds, Piece: 2319, Folio: 234, Page: 23
3 NOTE Secondary source - census enumeration
```

`buildPersonPropertyAssertionsIndex` was the only index feeding the exporter's assertion evidence and it read `assertion.Subject.Person` only, so `assertion.Subject.Event` was never consulted anywhere on the export path.

How: a new `buildEventPropertyAssertionsIndex` indexes assertions by `(eventID, property)`, and the `SOUR` is emitted under the event structure itself (`1 BIRT` / `2 SOUR`) rather than on the enclosing `INDI` — that is where GEDCOM 7 puts evidence for an event, and it keeps a birth's citation distinct from a death's. Family events (`MARR`, `DIV`) go through the same path, since `exportFamilyEvent` already had the event ID in hand; `exportPersonEvent` takes one now. References are collapsed by `(source XREF, PAGE)` before they are written, because a date assertion and a place assertion about one event routinely read off the same register entry and should not emit `SOUR @S1@` twice under one `DEAT`; evidence attached to the event entity itself is deduplicated against the assertion evidence in the same pass, and two different pages of one source stay two references. Assertions are indexed in sorted ID order so an event's subrecords are deterministic.

## Related issues

Fixes #1269

## Review focus

Two decisions the issue left open, both answered in the direction it suggested — worth a second opinion:

1. **Where the `SOUR` belongs.** Under the event structure rather than on the `INDI`. This matches GEDCOM 7's allowance for `SOUR` under an event and keeps evidence for a birth date out of a death record, at the cost of not sharing the person-level path.
2. **Deduplication key.** `(source XREF, PAGE)`, in the same spirit as #1120 on the import side. Two assertions citing one citation collapse to one `SOUR`; two pages of one source stay distinct. An assertion citing a source directly (no citation entity, so no `PAGE`) dedupes against another bare reference to that source but not against a paged one.

## Testing

- `make test` — full suite passes, including the existing GEDCOM export and round-trip tests.
- Seven new unit tests in `go-glx/gedcom_export_test.go` covering citations, direct sources, shared-page dedup, distinct-page retention, event-entity/assertion dedup, the family (`MARR`) path, and that an uncited event still exports bare.
- `glx/e2e/export_test.go` gains a case that exports `docs/examples/complete-family` through the real binary in both `-f 70` and `-f 551` and asserts the `SOUR` and `PAGE` lines under `BIRT` and `MARR`, plus the dedup count. It fails without the fix.
- `make check-links`, `make check-memory-drift`, `go vet` and `gofumpt -extra` clean on the changed files. `make lint` cannot run in the authoring environment — the installed `golangci-lint` is built with Go 1.25 and refuses the repo's `go 1.27.1` target — so `golangci-lint` in CI is the first full lint pass; its two findings (gofumpt's extra rules on a multi-line `append`, and `modernize` preferring `strings.SplitSeq`) are fixed here.

## Breaking changes

None to the archive format or the CLI. The GEDCOM an existing archive exports to gains `SOUR` subrecords under its event records where assertions provide them; unsourced archives are byte-identical. `exportPersonEvent` gains an `eventID` parameter, an unexported function in `go-glx`.


isaacschepp:
(withdrawn)

Issue #812, opened 2026-05-09
spec/2-core-concepts: evidence chain diagram ends at Property but should also show Event path


isaacschepp:
## Summary

`specification/2-core-concepts.md` (lines 787-792) shows the canonical evidence chain diagram ending at `Property`:

```text
Repository → Source → Citation → Assertion → Property
    ↓          ↓         ↓          ↓           ↓
 Physical   Original  Specific  Evidence-   Concluded
 Location   Material  Reference  Based      Value on
                                 Claim      Entity
```

This is the same bug as #571 (assertion-workflow README). After PR #360 (born/died removal), birth/death evidence flows to an `Event` (with `date`/`place` fields), not a Person property. The spec's canonical diagram should reflect that an Assertion can target either a Person/Place/Source property or an Event.

### Fix

Update the diagram to branch at the Assertion node, matching the model in `specification/4-entity-types/assertion.md`:

> The `subject` field uses a typed reference… Exactly one key must be present: `person`, `event`, `relationship`, or `place`.

The fix in #571 (PR landing the assertion-workflow README diagram update) provides a reference rendering.

### Related

- #571 — same bug in the assertion-workflow example README
- PR #360 — born/died removal (moved birth/death evidence to events)

Issue #713, opened 2026-04-20
glx stats: "Entity coverage" only counts direct assertion references, not transitive evidence coverage


isaacschepp:
## Area

cli

## Problem

`glx stats` displays an "Entity coverage (referenced by assertions)" block that's computed only from direct `assertion.subject.{person,event,relationship,place}` pointers. This misrepresents evidence coverage under the spec's own evidence-first model.

Example output from a freshly-captured archive (3 persons + 2 events + 3 relationships + 5 assertions, all sourced to a single christening citation):

````
Entity coverage (referenced by assertions):
  Persons         0/3  (0.0%)
  Events          2/2  (100.0%)
  Relationships   2/3  (66.7%)
  Places          0/5  (0.0%)
````

Every person is present on an event-participant list or on a relationship that IS referenced by an assertion. Every place is attached to an event that IS referenced by an assertion. Coverage is functionally 100% along the evidence chain — but the metric reads 0% because no assertion targets a person or place directly.

This is not a subtle edge case: the spec's own complete-family example attaches birth-date assertions to events (not persons) and place assertions to events. If a researcher follows the canonical evidence-first pattern, persons and places will always show 0% coverage until they add redundant person-level or place-level assertions purely to satisfy this counter.

## Source

`glx/stats_runner.go` lines 131–155. The four `covered*` maps are populated only from direct subject pointers; there is no traversal through events' `participants[].person`, events' `place`, or relationships' `participants[].person`.

````go
coveredPersons := make(map[string]struct{})
...
coveredPersons[a.Subject.Person] = struct{}{}
````

## Proposal

Pick one (or both):

(a) **Compute transitive coverage.** A person is "covered" if any assertion references them directly OR references an event/relationship they participate in. A place is "covered" if any assertion references it directly OR references an event that occurs there. This matches the user's intuition that a person is "sourced" when any claim about them has a citation.

(b) **Rename and re-scope.** Rename the existing block to "Direct assertion references" (making the narrow meaning explicit), and add a separate "Evidence coverage" block that does the transitive computation described in (a). Keep both for researchers who want to distinguish "facts directly asserted about the entity" from "entity appears anywhere in a sourced claim".

If keeping the current single-line-per-entity format, (b) is probably clearer than (a) because "coverage" is ambiguous enough that both metrics are useful.

## Context

[...]

PR #1291, opened 2026-09-19
feat: Let glx evidence take event, place and relationship subjects

### LINKED ISSUE #1268: glx evidence should accept event (and other) subjects, not persons only
### Area

cli

### Current limitation

`glx evidence <person> <property>` resolves its first argument to a person only. Pointing it at an
event is rejected:

```text
$ glx evidence event-death-1807 date
no person found matching "event-death-1807"
```

That is exactly as documented, so this is a scope request rather than a bug. The awkward part is
what it excludes. `glx evidence` is described as the tool for questions still in active research —
"it lays out the conflicting answers so you can weigh them" — but in an evidence-first archive the
two properties most likely to *have* conflicting answers, `date` and `place`, are normally asserted
on the **event**, not on the person.

That is the canonical pattern, not an unusual one. #713 makes the same observation from the other
side: "the spec's own complete-family example attaches birth-date assertions to events (not
persons) and place assertions to events." #812 is the spec-side version — the evidence chain
diagram stops at Property and never shows the Event path.

So an archive that follows the documented modelling gets a `glx evidence` that cannot be pointed at
its contested facts. Working around it means adding redundant person-level assertions purely to
make the command addressable, which is the same anti-incentive #713 describes for `glx stats`.

Concretely, on `main` @ `06a4dcb`, in a single-person archive with a birth date reckoned from a
death entry's decadal age band and cited to two reports, there is no way to ask the question the
command exists to answer:

```text
$ glx evidence person-hollnagel-michael-david date
# nothing — the date assertions are on event-birth-est / event-death-1807
```

### Proposed improvement

Let the subject argument be any assertable subject, matching what `add assertion` already accepts
(`--subject-person`, `--subject-event`, `--subject-place`, `--subject-relationship`):

```bash
glx evidence event-death-1807 date
glx evidence event-birth-est date
glx evidence place-liepen name
```

Resolution order could stay person-first to keep the name-search behaviour and avoid breaking the
current UX: try person (ID, then name), then fall through to an exact entity-ID match on the other
subject types. An ID with a known entity prefix could shortcut straight to the right type.

Two smaller things that would follow from it:

- The failure message should name what it accepts. `no person found matching "event-death-1807"`
  is accurate today but will be actively wrong once events resolve, and even now it reads as
  "that person does not exist" rather than "this command takes a person".
- A person-scoped roll-up (`glx evidence <person> date` gathering the date assertions from the
  events that person participates in) would be the more useful default for a researcher, but it is
  a bigger design question — whether the subject is "the person" or "the person's evidence chain" —
  so it is probably worth separating from the straightforward subject-type widening above.

### Context

Found while building a single-person archive from parish-register evidence, where every date and
place is event-subject and the birth date is explicitly a low-confidence inference from two
reports — precisely the case `glx evidence` is built to display.

Related: #713 (same modelling pattern, `glx stats`), #812 (spec diagram omits the Event path),
#1230 (`glx evidence` and assertion dates).

isaacschepp:
## What and why

Before: `glx evidence` resolved its first argument to a person, and nothing else. Pointing it at an event was rejected with `no person found matching "event-death-1807"`. In an evidence-first archive the two properties most likely to *have* conflicting answers — `date` and `place` — are asserted on the **event**, which is the modeling the specification's own complete-family example demonstrates. So an archive that followed the documented pattern had a command that could not be pointed at its contested facts, and the workaround was to add redundant person-level assertions purely to make the command addressable.

After: the subject argument accepts any assertable subject — person, event, place, or relationship — matching what `glx add assertion` already takes via `--subject-person`/`--subject-event`/`--subject-place`/`--subject-relationship`.

```text
$ glx evidence event-birth-john date --archive docs/examples/complete-family
Evidence for date of Birth (event-birth-john):
2 reports across 1 value

  1850-01-15 — 2 reports, best confidence: high
    citation-census-john             1851 Census of England and Wales high
    citation-john-birth              St. Paul's Parish Register - Leeds high

  Best evidence: 1850-01-15 (2 reports, high)
```

How: `findEvidenceSubject` replaces the `findPersonByQuery` call in `showEvidence`. It tries an exact entity ID across all four subject types first and only then the person name search, so an exact ID can never be shadowed by a substring name match, and it reports a query that is the ID of entities of more than one type rather than resolving it by map order. `matchingAssertions` and `collectEvidence` now take a `glxlib.EntityRef` and compare subjects by type and ID. `resolveAssertionValue` looks up the property definition in the vocabulary belonging to the subject's own entity type (`event_properties` for an event, and so on) rather than always in `person_properties`, and treats `place` on an event subject as a place reference, since that is the event's structural place field rather than a vocabulary property — the same mapping `glx migrate` uses.

Two things that follow from the widening:

- The failure message now names what the command accepts. `no person found matching "event-death-1807"` read as "that person does not exist" rather than "this command takes a person", and is simply wrong once events resolve. `findPersonByQuery` gains an `ErrNoPersonMatch` sentinel so `glx evidence` can recognize the no-match case and report its own wider vocabulary while keeping the ambiguous-name listing unchanged.
- `--format json` gains `subject`, `subject_type` and `subject_name`. The existing `person` and `person_name` keys are kept for person subjects so scripts written against the old shape keep working, and are omitted for every other subject type.

The person-scoped roll-up the issue raises as a second idea — `glx evidence <person> date` gathering the date assertions from the events that person participates in — is the larger design question the issue itself suggests separating, and is deliberately not part of this change.

## Related issues

Closes #1268

Related: #713 (same modeling pattern, `glx stats`), #812 (spec diagram omits the Event path).

## Review focus

- Resolution order. The issue suggested person-first (ID, then name) with a fall-through to other entity IDs; this does exact IDs across all four types first, then the person name search. An exact ID is the more specific match, so checking it first cannot shadow a name the caller meant, whereas the other order would let a substring name match swallow an ID naming a different entity. Worth a second opinion.
- The `--format json` compatibility aliases. Keeping `person`/`person_name` alongside `subject`/`subject_name` is the conservative choice; if breaking the old shape is acceptable pre-1.0, they can go.
- `place` on an event subject being special-cased as a place reference. It is a structural field, so no vocabulary definition declares it, but it is the property most likely to be looked at through this command.

## Testing

- `make test` — all packages pass.
- `make check-links`, `make check-memory-drift` — pass.
- `golangci-lint run ./...` at the pinned `v2.13.2` — one issue fewer than `main` (the `no person found matching` dynamic error became a wrapped sentinel); no new issues.
- New tests in `glx/evidence_runner_test.go`: every subject type resolving, the new error naming what it accepts, the ambiguous-name listing surviving, an ID shared across types being reported, subject labels and their ID fallbacks, event-subject collection, assertions not leaking across subject types, event `place` resolving to a place name, the subject-type vocabulary lookup, the text header for an event subject, the JSON key shape for both person and event subjects, and an end-to-end event-subject run.
- Manually against `docs/examples/complete-family`: `glx evidence event-birth-john date`, `glx evidence event-birth-john place`, `glx evidence "John" occupation`, and an unresolvable subject.

## Breaking changes

None. Person ID and person name lookups behave exactly as before, and the `--format json` keys that existed before this change are still emitted for person subjects.


isaacschepp:
Closing in favour of #1294, which carries the identical commit from `feat/evidence-any-subject`. This PR was opened from a `claude/…` branch, which `CLAUDE.md` disallows — branches use `feat/`/`fix/`/`docs/` prefixes. No review had happened here, so nothing is lost.

Issue #1272, opened 2026-09-18
Two tests fail on Windows on Unix-only filesystem assumptions


isaacschepp:
Two tests fail on a Windows checkout of `main`, both on assertions that only hold on a Unix filesystem. CI does not catch either — the Go jobs run on `ubuntu-latest` — so `go test ./glx/...` simply cannot go green for a contributor working on Windows.

Reproduced on `main` (`go1.27.1 windows/amd64`), with no local changes.

## 1. `TestSafeWriteMultiFileArchive/refuses_to_wipe_stale_backup_with_unrecovered_media_binaries`

```text
archive_io_test.go:1211: error should mention media/files; got: stale backup contains non-archive file
from a previous failed run; move or inspect it before retrying:
C:\...\001\archive.bak contains "C:\\...\\001\\archive.bak\\media\\files"
```

The guard fires correctly; only the assertion fails. The test asserts the message contains the literal `media/files`, while the message embeds a native path, so on Windows it reads `media\files`.

Worth deciding which side is wrong: the message currently interpolates an absolute native path, which is also why it is long and hard to read. Reporting the offending entry as a slash-separated path relative to the backup directory would fix the test and make the message better on every platform.

## 2. `TestPreflightFileOps_ReportsUnexpectedStatError`

```text
rename_runner_test.go:612: Error: An error is expected but got nil.
```

The test plants a regular file named `file` and then asks `preflightFileOps` about `file/x.glx`, relying on `Lstat` returning `ENOTDIR` for a path whose parent component is not a directory. Windows returns `ERROR_PATH_NOT_FOUND` there, which Go maps to `os.ErrNotExist` — exactly the "target does not exist, nothing to check" case `preflightFileOps` treats as success, so no error is returned.

The production behaviour looks defensible on both platforms (the create would fail later anyway); it is the test's way of provoking an unexpected `Lstat` error that is Unix-specific.

## Suggested fix

- (1) build the message from a slash-separated relative path — a fix in the production message, with the test left as-is;
- (2) either skip on `runtime.GOOS == "windows"` with a comment, or provoke the unexpected-stat error in a platform-neutral way.

## Why this matters

`CONTRIBUTING.md` asks contributors to run `make test` before opening a PR. On Windows the run always ends red, so the signal is useless there and a real regression is easy to wave past. Found while working on #91 (static-site follow-ups), unrelated to that work.

Issue #1293, opened 2026-09-19
ci: validate-examples reports a failure when its own concurrency group cancelled the run


isaacschepp:
### Area

tooling

### Summary

Every follow-up push to a PR branch leaves a spurious red `validate-examples` check behind on the superseded commit, and because `validate-examples` is a required status-check context the red X sits in the PR's checks list looking like a real failure.

The cause is an interaction between two changes that each landed correctly on their own:

- #346 added `concurrency: { group: validate-spec-…, cancel-in-progress: true }` to `validate-spec.yml`, so a second push cancels the in-flight run.
- #1035 fixed the aggregator's handling of a `skipped` matrix — but the guard it landed treats `cancelled` as a failure:

  ```yaml
  if [ "$RESULT" = "failure" ] || [ "$RESULT" = "cancelled" ]; then
    echo "::error::One or more example validations failed"
    exit 1
  fi
  ```

  (`.github/workflows/validate-spec.yml`, the `validate-examples` step)

`validate-examples` carries `if: always()`, so when the concurrency group cancels a run whose `validate-example` matrix legs are still queued, the matrix result is `cancelled`, the aggregator still runs, and it exits 1. The failure is reported against the old head SHA with the message "One or more example validations failed", which is untrue — nothing was validated and nothing failed.

Worth noting that #1035's own proposed fix is the code that is there now, so this is not a missed case from that issue so much as a consequence of it meeting #346. Distinguishing `cancelled` from `failure` is the remaining gap.

### Reproduction

1. Open a PR against `main` and let CI start.
2. Push a second commit while `validate-example` matrix legs are still queued or running.
3. The run for the first commit is cancelled by the concurrency group.
4. `validate-examples` on that first commit completes with `failure`.

Observed on #1281 — run [35407559440](https://github.com/genealogix/glx/actions/runs/35407559440), job `validate-examples`, against head `7bf1f2d`. The step log is unambiguous:

```text
build-cli result: success
Example validation matrix result: cancelled
##[error]One or more example validations failed
##[error]Process completed with exit code 1.
```

with `BUILD_RESULT: success` and `RESULT: cancelled` in the step env.

### Context

A genuine cancellation is not a validation failure and should not be reported as one. There are two distinct cases behind `cancelled` that are worth separating:

- **Superseded by the concurrency group** — the expected, benign case after any follow-up push. The run's verdict is irrelevant because a newer run supersedes it.
[...]

PR #1295, opened 2026-09-19
ci: Run the Go suite on Windows

isaacschepp:
## What and why

Before: the Go jobs in `validate-spec.yml` ran only on `ubuntu-latest`, so a Windows-only breakage reached contributors instead of CI. That is exactly how the two Unix-only test assumptions in #1272 went unnoticed, leaving the `make test` run `CONTRIBUTING.md` asks for permanently red on a Windows checkout.

After: a `test-windows` job runs the Go suite on `windows-latest`, so a Windows-only regression shows up on the PR that causes it.

> **Merge order.** This originally targeted #1281 and was retargeted to `main` when that merged. It now also carries the commits from #1301, because the new job found two *further* Windows-only failures on its first runs — see below. Merge #1301 first and those extra files drop out of this diff on the next base merge, leaving the workflow file and the changelog line.

## How

New `test-windows` job, `windows-latest`, `name: Go tests (Windows)`, running `go test ./glx/... ./go-glx/... ./tools/... -timeout 25m`.

Design decisions worth reviewing:

- **A separate job, not an `os` matrix on `test-conformance`.** A matrix renames that job's check contexts to `test-conformance (ubuntu-latest)`, and the bare `test-conformance` is a required context in the Main Protection ruleset — renaming it would leave the ruleset waiting on a check that never reports and deadlock merges. This is the hazard the workflow's own job-naming note warns about, so the new job stands alone.
- **Same fail-open `changes.outputs.go` gate** as the other Go jobs, written identically (`!cancelled() && (needs.changes.result != 'success' || …)`), so a docs-only PR skips it and a broken `changes` job runs it anyway. Windows runners bill at 2x Linux minutes, so the gate matters more here than elsewhere.
- **30-minute timeout** against `test-conformance`'s 20 — the same suite is appreciably slower on Windows, and the large GEDCOM fixtures dominate.
- **Narrower scope on purpose.** No coverage profile and no Codecov upload: the Linux run owns coverage, and a second unflagged report for the same commit would not be meaningful. No `-v`, which keeps the log readable. The stderr `Warning:` gate stays Linux-only because it tests a property of the code rather than of the platform, and running it in both places doubles the maintenance for no extra signal.
- **`shell: bash`** on the test step rather than the `windows-latest` default of `pwsh`, so a non-zero `go test` exit fails the step the same way it does everywhere else in this workflow.
- Both `uses:` lines reuse the SHA pins already in the file, per the repo-wide pinning rule in `SECURITY-POSTURE.md`.

## What the job found

The Review focus below asked whether the first run would be green, and promised to report anything it turned up. It turned up two more Windows-only failures in package `glx`, both test-setup bugs rather than product bugs, and both older than either PR:

1. `TestCacheGitOpTimeoutFallback` raced a one-nanosecond timeout against a real go-git repository read. Windows' coarser runtime timer let the read win, so a real commit SHA came back instead of the expected give-up. The two runs failed a different number of assertions, which is the timing-race signature.
2. `TestDefaultInitBranch_ReadsRealGitconfig` pointed `HOME` at a temp directory holding a stub gitconfig, but go-git finds that file through `os.UserHomeDir`, which reads `USERPROFILE` on Windows. The stub was never found and the lookup fell through to its `main` fallback.

Both are fixed in #1301, against `main`, since one of them touches production code this PR has no business changing. They are merged in here so the new job can be shown green before it lands.

## Manual ruleset change required

**This PR cannot make the new check required — that is a repository-settings change only a maintainer can make.** After merge, add this context to the Main Protection ruleset's required status checks:

```text
Go tests (Windows)
```

That string is the job's `name:`, which is what GitHub reports as the check context. If you would rather the context be the bare job key `test-windows`, drop the `name:` line instead and add `test-windows` — but pick one before making it required, because changing it afterwards deadlocks merges on a context that never reports.

Nothing else in the ruleset changes, and no existing required context is renamed or removed by this PR.

Until that change is made, a Windows-only regression shows as a red check that does not block merging. That is the intended first step: it makes the signal visible without risking a merge deadlock on a job whose Windows behaviour has never run in this repository before.

## Related issues

Refs #1272 — that issue is fixed by #1281; this PR is the CI coverage that keeps it fixed. Deliberately not a `Fixes`, since #1281 is what closes it.

## Review focus

- The job-vs-matrix decision and the required-context string, above. Those are the two choices that are expensive to change later.
- Whether `./tools/...` belongs in the Windows run. I included it to match `test-conformance`'s package set; dropping it would make the two jobs' scopes diverge.
- The two ported test fixes are #1301's to review; they are here only so this job can be verified.

## Testing

The workflow change cannot be exercised locally; this PR's own CI run is the test. Verified before pushing:

- The file parses as YAML and the job registers with the intended key, `runs-on` and `name` (`yaml.safe_load`, then inspecting `jobs['test-windows']`).
- Both `uses:` lines are full SHA pins with trailing version comments, matching the pins already used elsewhere in the file — no bare `@vN` introduced.
- `bash scripts/check-links.sh` (730 relative links OK) and `make check-memory-drift` for the `CHANGELOG.md` edit.
- The `go` paths filter already lists `.github/workflows/validate-spec.yml`, so this change triggers the Go jobs on its own PR without a filter edit.

## Breaking changes

None to the shipped tool. One new CI job, not required by the ruleset until a maintainer adds the context above, so no PR can be blocked by it on merge of this change.


isaacschepp:
Note for reviewers: `validate-examples` shows a red X on superseded heads of this branch (`f2ade47`, `4bf54b1`). It is not a failure of anything in this PR.

That job is an `if: always()` aggregator over `build-cli` and the `validate-example` matrix, and both of its guards treat a `cancelled` upstream job as a failure:

```bash
if [ "$BUILD_RESULT" != "success" ]; then   # trips when build-cli is cancelled
if [ "$RESULT" = "failure" ] || [ "$RESULT" = "cancelled" ]; then
```

Each of those runs was cancelled by the workflow's `cancel-in-progress` concurrency group when the next commit was pushed, so no example validation actually failed — the job logs record `BUILD_RESULT: success, RESULT: cancelled` on `f2ade47` and `BUILD_RESULT: cancelled, RESULT: cancelled` on `4bf54b1`. Filed separately as #1293; no fix for it exists yet, so there is nothing to port into this PR.

The current head has its own run, and nothing is red on it. Re-running a superseded commit's run buys nothing, since that commit is no longer the head.

noahbjohnson (dismissed):

isaacschepp:
The new `Go tests (Windows)` job found real Windows-only failures on its first outings, which is the job doing exactly what it was added for — but it means this PR cannot go green on its own, so the diff carries extra files for now.

**What failed**, both in package `glx` and both test-setup bugs rather than product bugs:

1. `TestCacheGitOpTimeoutFallback` sets `gitOpTimeout` to one nanosecond and expects `gitHeadSHA` to give up, which races a runtime timer against a real go-git repository read. Windows' timer is far coarser than Linux's, so the read regularly won and a real commit SHA came back. The two runs failed differently — three assertions in [run 35475741886](https://github.com/genealogix/glx/actions/runs/35475741886), one in [run 35476219282](https://github.com/genealogix/glx/actions/runs/35476219282) — the signature of a timing race rather than a deterministic break.
2. `TestDefaultInitBranch_ReadsRealGitconfig` points `HOME` at a temp directory holding a stub gitconfig, but go-git finds that file through `os.UserHomeDir`, which reads `USERPROFILE` on Windows. The stub was never found, so the branch lookup fell through to its `main` fallback and the assertion failed ([run 35800858803](https://github.com/genealogix/glx/actions/runs/35800858803)).

Neither is a regression from this PR or from #1281 — both tests have been wrong since they were written, and only ever ran on `ubuntu-latest`.

**What I did:** the fixes are their own PR against `main`, #1301, since one of them is production code this PR has no business touching. They are merged in here as well so the Windows job can be verified before this lands — those merges no-op once `main` carries #1301, and this PR's diff collapses back to the workflow file and the changelog line.

`main` is also merged in, resolving the earlier conflict.

isaacschepp:
Disregard this — no action needed.

PR #1301, opened 2026-09-23
fix: Make two glx tests pass on Windows

isaacschepp:
## What and why

Two tests in package `glx` failed on Windows and passed on Linux. Both are test-setup bugs rather than product bugs, and both went unnoticed because the Go jobs only ever ran on `ubuntu-latest`. The Windows job #1295 adds found them on its first outings; it cannot go green until they are fixed, so they are fixed here, separately from that PR.

### `TestCacheGitOpTimeoutFallback` — raced the clock

Before: it shrank `gitOpTimeout` to one nanosecond and expected `gitHeadSHA` and `gitWorkingTreeClean` to give up and return their "git unavailable" fallbacks. That set a real repository read racing against a runtime timer, and which of the two won depended on the platform's timer granularity rather than on the code under test. On Linux the timer fired first; on Windows, where the runtime timer is far coarser, the go-git read regularly finished first, `gitHeadSHA` returned a real commit SHA, and the test failed — with a different number of failing assertions from run to run.

After: the test's outcome does not depend on timing at all. It sets the deadline to zero, which is read as "already passed", so the lookups take the give-up path on every platform and every run. The deadline branch itself still gets exercised, by a new test whose operation blocks until the test releases it — that one cannot finish first however coarse the clock is, so a slow or loaded runner only makes the assertion wait longer, never flip.

No behaviour change for users: no production call site configures a non-positive timeout, and the 5s default is untouched.

### `TestDefaultInitBranch_ReadsRealGitconfig` — looked for `HOME` on Windows

Before: it wrote a gitconfig into a temp directory, pointed `HOME` at it, and asserted `defaultInitBranch` read `init.defaultBranch` back out. go-git locates that file with `os.UserHomeDir`, which reads `HOME` on Unix but `USERPROFILE` on Windows, so on Windows the stub was never found, the lookup fell through to the `main` fallback, and the assertion compared `refs/heads/main` against the expected `refs/heads/heritage`.

After: the test sets both variables, so the stub is found on every platform. One line, test-only.

## How

`gitHeadSHA` and `gitWorkingTreeClean` each carried their own copy of the same goroutine-and-`select` boilerplate. That is now a single generic helper, `boundedGitOp`, which runs the operation under `gitOpTimeout` and returns the zero value and `false` when it misses the deadline. `boundedGitOp` treats a non-positive `gitOpTimeout` as a deadline that has already elapsed and abandons the operation before starting it — the same thing `context.WithTimeout(ctx, 0)` does, and the seam the test now uses. The comment on `gitOpTimeout` that explained the abandoned-goroutine mechanics moves onto the helper, where the mechanics now live.

`boundedGitOp` carries a `//nolint:ireturn`: the linter reads its type-parameter return as an interface return, but a type parameter is not an interface — each caller gets its own concrete type back. A targeted directive seemed better than adding `generic` to the linter's allow list repo-wide.

## Related issues

Refs #1272 — that issue covered the two Unix-only assumptions fixed in #1281. These are the third and fourth Windows-only failures, surfaced by the job #1295 adds. Neither is a regression from either PR: both tests have been wrong since they were written.

## Review focus

Correctness of the `boundedGitOp` extraction — that the refactor preserves the abandoned-goroutine guarantees the old comment documented (buffered capacity-1 channel, result passed through the channel rather than a shared variable, no leak and no data race with a late-finishing goroutine), and that the non-positive-timeout shortcut is an acceptable way to expose the give-up path to tests. The `init_git_test.go` change is a one-liner.

## Testing

- `go test ./glx/ -run 'TestCacheGit|TestBoundedGitOp' -count=5` and `-race -count=2` — all pass.
- `go test ./glx/ -run 'TestDefaultInitBranch|TestInitArchiveGitRepo' -count=2` — passes.
- Full local suite `go test ./glx/... ./go-glx/... ./tools/...` — green.
- `go vet ./glx/...`, `gofmt` clean on the changed files, `go build ./...`.
- Coverage on the changed code: `boundedGitOp` 100%, `gitHeadSHA` 100%, `gitWorkingTreeClean` 88.2% (its remaining gaps are pre-existing error branches).
- Windows itself is verified by the new job in #1295, which merges these commits in so it can go green before it lands.

Note for a maintainer, unrelated to this diff: `gofmt -l glx/` also reports `glx/archive_io_test.go` on current `main` (map-literal alignment in a foreign-files fixture). Left alone here rather than widening this PR.

## Breaking changes

None.


isaacschepp:
Disregard this — no action needed.

